PR #32, opened 2024-05-10
Getting build urls for when uuid and baseline are set

paigerube14:
## Type of change

- [ ] Refactor
- [ ] New feature
- [X] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
With the changes from https://github.com/cloud-bulldozer/orion/pull/15, when using the uuid and baseline parameters I had added, it was now throwing an error trying to find the buildUrl for each uuid. This will get that data without erroring 

## Related Tickets & Documents

- Related Issue #
- Closes #

## Checklist before requesting a review

- [X] I have performed a self-review of my code.
- [ ] If it is a core feature, I have added thorough tests.

## Testing
- Please describe the System Under Test.
- Please provide detailed steps to perform tests related to this code change.
- How were the fix/results from this change verified? Please provide relevant screenshots or results.
`$orion --uuid $UUID --baseline $BASELINE_UUID --config examples/small-scale-node-density-cni.yaml --debug`
```
,uuid,timestamp,P99,apiserverCPU_cpu_avg,ovnCPU_cpu_avg,etcdCPU_cpu_avg,etcdDisk_duration_avg,buildUrl
0,26feeb47-3f7d-45b4-9926-f8f774a08385,2023-12-14T23:13:52.563810987Z,2000,22.754171747822074,3.4660279278374864,21.24262421293023,0.010324644209403131,https://tinyurl.com/26843gjw
1,0f731581-6630-4620-921f-3d9fa11d1330,2024-05-07T21:20:02.737893524Z,2000,19.23959183682067,3.21447647319984,16.56003964168054,0.009221293420220414,https://tinyurl.com/266dh3ay
```

PR #33, opened 2024-05-12
Consider all the iterations for control plane test comparisons

vishnuchalla:
## Type of change

- [x] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Prerequisite: https://github.com/cloud-bulldozer/orion/pull/32

To consider all the iterations while comparing control plane tests. Because iterations are calculated in the run-time based on ppn for all the node-density tests so that they can be ignored in the fingerprint. Please feel free to correct me If I am wrong.

## Testing
Tested and verified in local able to fetch all the similar UUIDs.

### DIFF
--- a/orion.py
+++ b/orion.py
@@ -67,7 +67,7 @@ def orion(**kwargs):
         benchmarkIndex=test['benchmarkIndex']
         uuid = kwargs["uuid"]
         baseline = kwargs["baseline"]
-        match = Matcher(index="ospst-perf-scale-ci-*",
+        match = Matcher(index="ospst-perf-scale-ci*",
                         level=level, ES_URL=ES_URL, verify_certs=False)
         if uuid == "":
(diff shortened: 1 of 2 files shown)


orion.py
```diff
@@ -67,7 +67,7 @@ def orion(**kwargs):
         benchmarkIndex=test['benchmarkIndex']
         uuid = kwargs["uuid"]
         baseline = kwargs["baseline"]
-        match = Matcher(index="ospst-perf-scale-ci-*",
+        match = Matcher(index="ospst-perf-scale-ci*",
```
paigerube14, line 70: this was added to the config files themselves. We need to update to get the index from the file

vishnuchalla:
Closing this to have an action item to populate pods per node value in the run fingerprint (i.e published as a part of index.sh from prow) to match similar node-density runs.

PR #79, opened 2024-10-30
Adding additional metrics to track

jtaleric:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Adding multus/monitoring

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/examples/trt-external-payload-cluster-density.yaml
+++ b/examples/trt-external-payload-cluster-density.yaml
@@ -35,6 +35,28 @@ tests :
         - "[Jira: kube-apiserver]"
       direction: 1
 
+    - name: multusCPU
+      metricName : containerCPU
+      labels.namespace.keyword: openshift-multus
+      metric_of_interest: value
+      agg:
+        value: cpu
+        agg_type: avg
+      labels:
+        - "[Jira: multus]"
+      direction: 1
+
+    - name: monitoringCPU
+      metricName : containerCPU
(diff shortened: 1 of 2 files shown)


afcollins:
Do we also want these on node-density or only care about them from cluster-density?

examples/trt-external-payload-cluster-density.yaml
```diff
@@ -35,6 +35,28 @@ tests :
         - "[Jira: kube-apiserver]"
       direction: 1
 
+    - name: multusCPU
```
afcollins, line 38: We should decide whether we want all metrics (including those not added here) to aggregate only the workload or also include the garbage collection metrics (lines 21-22 above).

examples/trt-external-payload-cluster-density.yaml
```diff
@@ -35,6 +35,28 @@ tests :
...
+      metricName : containerCPU
+      labels.namespace.keyword: openshift-multus
+      metric_of_interest: value
+      agg:
+        value: cpu
+        agg_type: avg
```
afcollins, line 44: This averages all pods and containers in the namespace .. do we want to focus on a container specifically? I think kube-rbac-proxy would be weighing it down

afcollins (approved):

Issue #78, opened 2024-10-10
Orion to not alert already detected regressions


shashank-boyapally:
Orion has now fails for every regression the idea is to fail only for new unknown regressions

rsevilla87:
closed in https://github.com/cloud-bulldozer/orion/pull/80

PR #80, opened 2024-12-16
Adding method to ack

jtaleric:
## Type of change
We need a mechanism to ack issues we have opened issues for, or issues we have brought to the attention of dev teams.

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

- Related Issue #
- Closes #

## Testing


README.md
```diff
@@ -163,6 +163,18 @@ You can open the match requirement by using the ```--node-count``` option to fin
 
 **_NOTE:_**  The ```cmr```, ```--hunter-analyze``` and ```--anomaly-detection``` flags are mutually exclusive. They cannot be used together because they represent different algorithms designed for distinct use cases.
 
+#### Ack known bugs
+To ack known regressions, you must provide a yaml file with the timestamp which the regression was identified at.
+
+```
+---
+ack :
+  - timestamp: 1733490603,
```
josecastillolema, line 172: So you can ack based on timestamp or uuid?

josecastillolema:
Closes https://github.com/cloud-bulldozer/orion/issues/78

josecastillolema:
So the **_ack.yml_** file will live in the orion repository right?
How often do you envision this file will need to be updated?

vishnuchalla:
This approach will bring us overhead to specify yaml manually to acknowledge regressions which are just being deleted from our list of changepoints. Instead we need to think of an automated way of acknowledging regressions so that there is no manual intervention. 

One possible idea could be, Why not have orion publish all the regressions detected to a separate elastic search index automatically? it can query list of existing regressions from the persistent store (.i.e es index that it has) and delete them from its reporting list to avoid duplicates every single time it notifies regressions.

jtaleric:
> This approach will bring us overhead to specify yaml manually to acknowledge regressions which are just being deleted from our list of changepoints. Instead we need to think of an automated way of acknowledging regressions so that there is no manual intervention. 

IMHO Performance Regressions *should* be manually ack'ed. 

I don't think automated acks are the right way. 


> 
> 
> 
> One possible idea could be, Why not have orion publish all the regressions detected to a separate elastic search index automatically? it can query list of existing regressions from the persistent store (.i.e es index that it has) and delete them from its reporting list to avoid duplicates every single time it notifies regressions.

This sounds like more overhead and additional accounting due to yet another index in OpenSearch. Today Orion is purely RO when it comes to OpenSearch.

jtaleric:
> So the **_ack.yml_** file will live in the orion repository right?
> 
> How often do you envision this file will need to be updated?

The workflow I proposed to @shashank-boyapally : 

1. Regression is found
2. PerfEng User engages with component team, determines to open a bug 
3. PerfEng who engaged with component team opens PR to add the ack. 

This way we know who is acking in the event we need to look back and ask questions.

vishnuchalla:
> > So the **_ack.yml_** file will live in the orion repository right?
> > How often do you envision this file will need to be updated?
> 
> The workflow I proposed to @shashank-boyapally :
> 
> 1. Regression is found
> 2. PerfEng User engages with component team, determines to open a bug
> 3. PerfEng who engaged with component team opens PR to add the ack.
> 
> This way we know who is acking in the event we need to look back and ask questions.

So, do you mean to say that we will have a yaml in this repo itself which will be updated with the list of regressions already detected?

jtaleric:
> > > So the **_ack.yml_** file will live in the orion repository right?
> > > How often do you envision this file will need to be updated?
> > 
> > 
> > The workflow I proposed to @shashank-boyapally :
> > 
> > 1. Regression is found
> > 2. PerfEng User engages with component team, determines to open a bug
> > 3. PerfEng who engaged with component team opens PR to add the ack.
> > 
> > This way we know who is acking in the event we need to look back and ask questions.
> 
> So, do you mean to say that we will have a yaml in this repo itself which will keep on updating manually with the list of regressions every single time they are detected?

It doesn't need to be in this repo, but yes, the workflow should take a manual effort.. Regressions shouldn't be automagically silenced, IMHO.

README.md
```diff
@@ -163,6 +163,18 @@ You can open the match requirement by using the ```--node-count``` option to fin
...
+    metric: "etcdCPU_avg"
```
josecastillolema, line 173: Should we add a reason/description file?
A relatively small string to link some jiras, describe the issue, etc.?

pkg/algorithms/edivisive/edivisive.py
```diff
@@ -16,15 +16,34 @@ class EDivisive(Algorithm):
...
+            for ack in self.options["ackMap"]["ack"]:
+                pos = series.find_by_attribute("uuid",ack["uuid"])
+                ackList.append(
+                    {"pos" : pos[0],
+                     "metric" : ack["metric"]})
+
         # filter by direction
         for metric, changepoint_list in change_points_by_metric.items():
             for i in range(len(changepoint_list)-1, -1, -1):
                 if ((self.metrics_config[metric]["direction"] == 1 and changepoint_list[i].stats.mean_1 > changepoint_list[i].stats.mean_2) or
```
vishnuchalla, line 35:
```suggestion
                if ((self.metrics_config[metric]["direction"] == 1 and changepoint_list[i].stats.mean_1 > changepoint_list[i].stats.mean_2) or (self.metrics_config[metric]["direction"] == -1 and changepoint_list[i].stats.mean_1 < changepoint_list[i].stats.mean_2) or (str(changepoint_list[i]) + "_" + changepoint_list[i].metric in ackSet)):
```

pkg/algorithms/edivisive/edivisive.py
```diff
@@ -16,15 +16,34 @@ class EDivisive(Algorithm):
...
                     (self.metrics_config[metric]["direction"] == -1 and changepoint_list[i].stats.mean_1 < changepoint_list[i].stats.mean_2) ):
                     del changepoint_list[i]
+
+        # Filter ack'ed changes
```
vishnuchalla, line 39: This won't be required with the above logic that uses a set.

pkg/algorithms/edivisive/edivisive.py
```diff
@@ -16,15 +16,34 @@ class EDivisive(Algorithm):
...
+        # Process if we have ack'ed regression
+        ackList = []
+        if len(self.options["ack"]) > 1 :
+            for ack in self.options["ackMap"]["ack"]:
+                pos = series.find_by_attribute("uuid",ack["uuid"])
+                ackList.append(
+                    {"pos" : pos[0],
+                     "metric" : ack["metric"]})
+
         # filter by direction
```
vishnuchalla, line 32:
```suggestion
        # filter by direction and already ack' ed changes
```

pkg/algorithms/edivisive/edivisive.py
```diff
@@ -16,15 +16,34 @@ class EDivisive(Algorithm):
     def _analyze(self):
         self.dataframe["timestamp"] = pd.to_datetime(self.dataframe["timestamp"])
         self.dataframe["timestamp"] = self.dataframe["timestamp"].astype(int) // 10**9
-        series= self.setup_series()
+        series = self.setup_series()
         change_points_by_metric = series.analyze().change_points
 
+
+        # Process if we have ack'ed regression
+        ackList = []
```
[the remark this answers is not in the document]
jtaleric: hm. Not sure I like the idea of string parsing, because if we have a component that ever has a `_`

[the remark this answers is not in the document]
vishnuchalla: Regardless of the characters that a component has in its name, the join for index_componentName will always be unique within this code block (.i.e for example 0_etch_avg_cpu, 0__etch_avg_memory etc). Its the same idea previously used in map being represented differently as a string by just using `_` as a delimiter so that we can avoid on redundant loops to optimize time complexity.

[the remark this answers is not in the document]
jtaleric: I think I understand better! Thanks!

README.md
jtaleric: at first I thought to use the timestamp, but UUID is the path I went with.

vishnuchalla (approved):
lgtm

PR #88, opened 2025-01-21
Remove unused config files

vishnuchalla:
## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description
Removing unused configs that are already a part of the examples folder. Also didn't notice them being used anywhere else.

## Testing
CI tests in this PR should go through.


jtaleric:
What do you mean by unused? I use the `trt-payload-cluster-density`

vishnuchalla:
> What do you mean by unused? I use the `trt-payload-cluster-density`

We have the same config file in [examples folder](https://github.com/cloud-bulldozer/orion/blob/main/examples/trt-payload-cluster-density.yaml) as well. And the same applies to other files in `configs` folder. Can we just keep all the configurations in one folder?

jtaleric:
> > What do you mean by unused? I use the `trt-payload-cluster-density`
> 
> We have the same config file in [examples folder](https://github.com/cloud-bulldozer/orion/blob/main/examples/trt-payload-cluster-density.yaml) as well. And the same applies to other files in `configs` folder. Can we just keep all the configurations in one folder?

ah, good point. yes.

jtaleric (approved):

PR #89, opened 2025-01-21
CI fix to always capture latest ocp version

vishnuchalla:
## Type of change

- [x] Refactor
- [x] New feature
- [x] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description
Code changes to fetch latest OCP version from our CI docs published in the elastic search. That way we will always end up querying docs that are actually present as a part of our CI.

- Related Issue #
- Closes #

## Testing
Verified in local by running the same set of commands as added in this PR.
```
echo $version
4.17
```
Also PR's CI tests went through.


vishnuchalla:
@josecastillolema This should fix issues with your [PR](https://github.com/cloud-bulldozer/orion/pull/87). PTAL and let me know.

examples/payload-scale-416.yaml
```diff
@@ -10,7 +10,7 @@ tests :
       workerNodesType: m6a.xlarge
       workerNodesCount: 6
       benchmark.keyword: cluster-density-v2
-      ocpVersion: 4.16
+      ocpVersion: {{ version }}
```
paigerube14, line 13: is this file and the payload-scale-415.yaml now the same? Are we able to combine them and remove one of them

paigerube14:
Are we able to edit the names of the tests here because they are now not actually running on 4.15 and 4.16, they'll run on the latest version 

https://github.com/cloud-bulldozer/orion/blob/main/test.bats#L36-L42

examples/readout-control-plane-node-density.yaml
```diff
@@ -3,7 +3,7 @@ tests :
     index: {{ es_metadata_index }}
     benchmarkIndex: {{ es_benchmark_index }}
     metadata:
-      ocpVersion: 4.15
```
paigerube14, line 6: there are multiple other places version is specified in this file, can you make sure to edit all of them

examples/readout-control-plane-cdv2.yaml
```diff
@@ -3,7 +3,7 @@ tests :
     index: {{ es_metadata_index }}
     benchmarkIndex: {{ es_benchmark_index }}
     metadata:
-      ocpVersion: 4.15
```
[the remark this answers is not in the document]
vishnuchalla: Closing as duplicate.

examples/readout-control-plane-node-density.yaml
vishnuchalla: Updated as suggested.

examples/payload-scale-416.yaml
vishnuchalla: Yes, removed the redundant ones.

vishnuchalla:
> Are we able to edit the names of the tests here because they are now not actually running on 4.15 and 4.16, they'll run on the latest version
> 
> https://github.com/cloud-bulldozer/orion/blob/main/test.bats#L36-L42

Updated. Also once we merge this PR, before cutting a release we need to remove references [like this](https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift-qe/orion/openshift-qe-orion-ref.yaml#L30) in prow

jtaleric (approved):
/lgtm

paigerube14 (approved):
/lgtm

jtaleric:
> > Are we able to edit the names of the tests here because they are now not actually running on 4.15 and 4.16, they'll run on the latest version
> > https://github.com/cloud-bulldozer/orion/blob/main/test.bats#L36-L42
> 
> Updated. Also once we merge this PR, before cutting a release we need to remove references [like this](https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift-qe/orion/openshift-qe-orion-ref.yaml#L30) in prow

right, we will have to use this method - > https://github.com/openshift/release/pull/56698

PR #95, opened 2025-02-10
add config for payload node-density-cni job

mohit-sheth:
## Type of change

- [ ] Optimization

## Description

Adding node-density-cni config

### DIFF
--- a/examples/trt-external-payload-node-density-cni.yaml
+++ b/examples/trt-external-payload-node-density-cni.yaml
@@ -1,0 +1,252 @@
+tests :
+  - name : payload-node-density-cni
+    index: {{ es_metadata_index }}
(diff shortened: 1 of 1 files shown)


examples/trt-external-payload-node-density-cni.yaml
```diff
@@ -0,0 +1,101 @@
...
+    - name: kubelet 
+      metricName : kubeletCPU 
+      metric_of_interest: value 
+      labels: 
+        - "[Jira: Node]" 
+      agg: 
+        value: cpu 
+        agg_type: avg
+      direction: 1
+
```
jtaleric, line 101: Do we want to track more OVN specific metrics here?

jtaleric (approved):
lgtm

PR #115, opened 2025-04-03
ack for ovn PR 2501 regression in ovnk CPU

mohit-sheth:
add an ack for https://issues.redhat.com/browse/OCPBUGS-54508


jtaleric (approved):
/lgtm

PR #128, opened 2025-04-24
Fmatch 0.1.0

rsevilla87:
## Type of change

- [x] Optimization

## Description

Filtering out okd results to prevent mixing up its results with old runs in openshift
PTAL @josecastillolema 

- Related Issue #
- Closes #


vishnuchalla (approved):
Thanks for this change. Definitely needed.

PR #160, opened 2025-07-25
make jobtype a var, defaults to periodic

mohit-sheth:
For jobs triggered by pull requests, we want to compare them against both periodic runs and previous PR runs to check for improvements.
In contrast, for CI (periodic) runs, we exclude PR runs from comparison to maintain consistency with our baseline.


jtaleric:
nice! Thanks!

PR #163, opened 2025-07-28
Adding orion examples for each chaos scenario

shahsahil264:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

- Related Issue #
- Closes #

## Testing


examples/pod_disruption_scenarios.yaml
```diff
@@ -0,0 +1,469 @@
...
+    uuid_field: run_uuid
+    metadata:
+      cluster_version: "{{ version }}"
+      scenarios.scenario_type: pod_disruption_scenarios
+      cloud_infrastructure: AWS  # Options: AWS, IBMCloud, GCP, Azure, VSphere
+      cloud_type: self-managed   # Options: self-managed, rosa
+      total_node_count: 9
+      node_summary_infos.instance_type: "*xlarge*"  # Matches: m6a.xlarge (masters), r5.xlarge (workers)
+      network_plugins: OVNKubernetes
+      scenarios.scenario: "*pod_scenario.yaml"
```
paigerube14, line 15: since we have the scenario_type already, think this variable is over kill

pkg/algorithms/algorithmFactory.py
```diff
@@ -12,7 +12,7 @@
 class AlgorithmFactory: # pylint: disable= too-few-public-methods, too-many-arguments, line-too-long
     """Algorithm Factory to choose algorithm
     """
-    def instantiate_algorithm(self, algorithm: str, matcher: Matcher, dataframe:pd.DataFrame, test: dict, options: dict, metrics_config: dict[str,dict]):
+    def instantiate_algorithm(self, algorithm: str, matcher: Matcher, dataframe:pd.DataFrame, test: dict, options: dict, metrics_config: dict[str,dict], version_field: str = "ocpVersion"):
```
paigerube14, line 15: is this a rebase issue or did you make these changes?

examples/pod_disruption_scenarios.yaml
shahsahil264: Yes, true, I added it, as we need it for the hog scenarios, as all for all hog scenarios, the scenarios.scenario_type is hog_scenarios, hence we need to use scenarios.scenario. We can definitely remove them from the other files

pkg/algorithms/algorithmFactory.py
shahsahil264: Had to make the changes, as I think orion by default expects the field for the ocp version to be ocpVersion, whereas we name it as cluster_version in our data

paigerube14:
one other thought here, could we make one generic file of all the metrics and let the environment variable set the scenario so we don't need a file per scenario when they are really all the same? 

this would exclude pod scenario and node scenarios since they have scenario specific metrics already (and maybe application if we are tracking the downtime)

shahsahil264:
> one other thought here, could we make one generic file of all the metrics and let the environment variable set the scenario so we don't need a file per scenario when they are really all the same?
> 
> this would exclude pod scenario and node scenarios since they have scenario specific metrics already (and maybe application if we are tracking the downtime)

Yes, its a great idea, I have made the changes, I'll have a commit up soon

examples/application_outages_scenarios.yaml
```diff
@@ -0,0 +1,501 @@
...
+      cloud_type: self-managed   # Options: self-managed, rosa
+      total_node_count: 9
+      node_summary_infos.instance_type: "*xlarge*"  # Matches: m6a.xlarge (masters), r5.xlarge (workers)
+      network_plugins: OVNKubernetes
+      scenarios.scenario: "*app_outages.yaml"
+    metrics:
+      # Recovery & Health metrics (specific to application outages scenarios)
+      - name: health_check_recovery
+        metricName: health_check_recovery
+        metric_of_interest: duration
```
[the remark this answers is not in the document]
shahsahil264: I think sum would be a good metric for that, by that I mean sum of all the status codes which are not 200

examples/chaos_results.yaml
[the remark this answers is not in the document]
shahsahil264: Thanks!

PR #173, opened 2025-08-06
Remove fmatch dependency

rsevilla87:
## Type of change

- [x] Optimization

## Description

Remove fmatch dependency and rename pkg to orion.

- Related Issue #
- Closes #


vishnuchalla (approved):
lgtm, because fmatch is only being used in orion at the moment.

PR #175, opened 2025-08-12
Remove daemon mode

rsevilla87:
## Type of change

- [x] Optimization
- [ ] Documentation Update

## Description

Removing daemon mode from orion.

Follow-up of #173


jtaleric (commented):
can you remove the `cmd` aspect since we will only have a single mode of running?

rsevilla87:
> can you remove the `cmd` aspect since we will only have a single mode of running?

done

PR #177, opened 2025-08-14
Update tests

paigerube14:
## Type of change

- [ ] Refactor
- [ ] New feature
- [X] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
The current tests are actually failing in a lot of cases because of the lack of data. I noticed the latest version that we get in the setup was getting an OKD version and didn't have much data. 

Want to also update the tests to fail when there is a lack of data since its not actually testing the command line arguments if no data is found  

Failure run when not enough data error returns 1 instead of 0: https://github.com/cloud-bulldozer/orion/actions/runs/16969211330/job/48101321838

- Related Issue #
- Closes #

- [] I have performed a self-review of my code.
- [X] If it is a core feature, I have added thorough tests.

## Testing

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -39,7 +39,23 @@ jobs:
 
     - name: Execute Tests
       run: |
-        bats -F pretty -T --print-output-on-failure test.bats
+        bats -F pretty -T --print-output-on-failure test.bats --gather-test-outputs-in outputs
       env:
         TERM: linux      
         QE_ES_SERVER: ${{ secrets.QE_ES_SERVER }}
+    
+    - name: Upload csv data
+      if: ${{ success() || failure() }}
+      uses: actions/upload-artifact@v4
+      with:
+        name: csvs
+        path: "*.csv"
+        if-no-files-found: error
+      
+    - name: Upload outputs data
+      if: ${{ success() || failure() }}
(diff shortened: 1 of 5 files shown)


josecastillolema:
About the OKD part, which example file are the test using?
They should have something like:
```
tests:
  - name: test1
    metadata:
      not:
        stream: okd
```

paigerube14:
> About the OKD part, which example file are the test using? They should have something like:
> 
> ```
> tests:
>   - name: test1
>     metadata:
>       not:
>         stream: okd
> ```

It was getting the latest version as 4.20.0-0.okd and not all tests had runs with that version

See change: https://github.com/cloud-bulldozer/orion/pull/177/files#diff-dabae2e005bdb42d0d4196deb8d65c7795812bda4a77aa41e296f7d899ac7496R50

paigerube14:
Added some changes that are also in https://github.com/cloud-bulldozer/orion/pull/163 to try to get the tests passing to merge

paigerube14:
@jtaleric @vishnuchalla any thoughts on this PR? I think this is super important as I had found an error in our chaos config in the test.bats file because it was reporting a false positive. It couldn't find data because of the incorrect value but was showing pass

jtaleric (approved):
/lgtm

PR #193, opened 2025-09-03
add instantenous ovsVswitchd CPU and Mem to track

mohit-sheth:
Make this as one of the metrics we track regularly

### DIFF
--- a/examples/small-scale-cluster-density.yaml
+++ b/examples/small-scale-cluster-density.yaml
@@ -54,3 +54,43 @@ tests :
         value: duration
         agg_type: avg
 
+    - name:  ovsCPU-irate-all
+      metricName.keyword: cgroupCPU
+      labels.id.keyword: /system.slice/ovs-vswitchd.service
+      metric_of_interest: value
+      agg:
+        value: cpu
+        agg_type: avg
+      direction: 1
+      threshold: 10
+
+    - name:  ovsMemory-Workers
+      metricName.keyword : cgroupMemoryRSS-Workers
+      labels.id.keyword : /system.slice/ovs-vswitchd.service
+      metric_of_interest: value
+      agg:
+        value: mem
+        agg_type: max
+      direction: 1
+      threshold: 10
+
+    - name:  ovsMemory-Masters
+      metricName.keyword : cgroupMemoryRSS-Masters
+      labels.id.keyword : /system.slice/ovs-vswitchd.service
+      metric_of_interest: value
+      agg:
+        value: mem
+        agg_type: max
+      direction: 1
+      threshold: 10
+
+    - name:  ovsMemory-all
+      metricName.keyword: cgroupMemoryRSS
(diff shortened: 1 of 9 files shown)


orion/algorithms/isolationforest/isolationForest.py
```diff
@@ -33,6 +34,11 @@ def _analyze(self):
         logger.info("Starting analysis using Isolation Forest")
         metric_columns = self.metrics_config.keys()
         dataframe_with_metrics = dataframe[metric_columns]
+        # Clean and impute: replace inf with NaN, then fill NaNs with column median, then 0 if still NaN
+        dataframe_with_metrics = dataframe_with_metrics.replace([np.inf, -np.inf], pd.NA)
```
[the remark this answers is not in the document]
mohit-sheth: however we should put a warning in case of NaN instead of exiting

jechen0648:
@vishnuchalla @mohit-sheth   what would it take to get this PR merge?

orion/algorithms/isolationforest/isolationForest.py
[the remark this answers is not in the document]
afcollins: I agree that NaN is a known error that orion should handle instead of allowing it to interrupt the run. It can log an error/warning and maybe even remove the dataframe since it won't be calculated accurately.

examples/small-scale-cluster-density.yaml
```diff
@@ -54,3 +54,44 @@ tests :
...
+      agg:
+        value: cpu
+        agg_type: avg
+      direction: 1
+      threshold: 10
+
+    - name:  ovsMemory-Workers
+      metricName.keyword : cgroupMemoryRSS-Workers
+      labels.id.keyword : /system.slice/ovs-vswitchd.service
+      labels.id.keyword: /system.slice/ovs-vswitchd.service
```
afcollins, line 70: dupe line here.

examples/small-scale-udn-l2.yaml
```diff
@@ -30,8 +30,8 @@ tests :
       direction: 1
       threshold: 10
 
-    - name:  ovsCPU
-      metricName : cgroupCPUSeconds-Workers
+    - name:  ovsCPU-Workers
+      metricName.keyword: cgroupCPUSeconds-Workers
```
afcollins, line 34: The agg_type here is avg. We should only have a single datapoint returned, but it should be 'max' to be consistent.

examples/small-scale-udn-l3.yaml
```diff
@@ -30,8 +30,8 @@ tests :
       direction: 1
       threshold: 10
 
-    - name:  ovsCPU
-      metricName : cgroupCPUSeconds-Workers
+    - name:  ovsCPU-Workers
+      metricName.keyword : cgroupCPUSeconds-Workers
```
afcollins, line 34: same comment as above in udn-l2

afcollins (commented):
Only one dupe line, but otherwise LGTM

PR #196, opened 2025-09-09
Config that attempts to look at all namespaces

jtaleric:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Trying to collect all the namespaces and ensure we don't have regressions against any of them. This is only looking at the CPU metric, we should also consider creating one for memory


mohit-sheth (approved):
lgtm

PR #219, opened 2025-10-22
Support to display mutiple fields

rsevilla87:
## Type of change

- [x] New feature

## Description

Adding support to display multiple fields. Also removing hardcoded `row.append(record["buildUrl"])`, as this column can be removed from the output

- Related Issue #
- Closes #


jtaleric:
@rsevilla87 still relevant?

PR #286, opened 2026-02-18
Windowing issue

Balatripura587:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

Changepoint detection (EDivisive/hunter) needs enough samples before and after a point to be reliable. We were sometimes reporting changepoints in the first few points (e.g. the 2nd) without that context, so anomalies were treated as regressions and caused noise in CI (e.g. OVN).

### Solution

- **Buffer:** The first 5 points are treated as the window edge (`CHANGEPOINT_BUFFER` in `orion/constants.py`). Any changepoint in that range is re-validated.
- **Window expansion:** If a changepoint is in the buffer, we expand the lookback window, then re-run detection. We keep it as a regression only if the expanded run still has a changepoint; otherwise we treat it as a false positive

### Testing

Verified with all three output formats (text, JSON, junit).
Example: 
`orion --lookback 15d --since 2026-01-20 --hunter-analyze --config hack/ci-tests/ci-tests-early-cp.yaml --metadata-index "orion-integration-test-data*" --benchmark-index "orion-integration-test-metrics*" --es-server=${QE_ES_SERVER} --node-count true --input-vars='{"version": "4.20"}' --output-format junit > results-early-cp.x
ml`
Output: 
Screenshot 2026-02-26 at 4 57 55 PM


chentex (approved):
/lgtm thanks @Balatripura587

Issue #261, opened 2026-01-19
Automatically load ACK files for known regressions


mohit-sheth:
## Summary

  Currently, users must manually specify ACK files using the --ack flag to acknowledge known regressions. Since these ACK files represent documented, known regressions that should always be considered during analysis, they should be automatically loaded based on the test configuration.

  #### Current Behavior

  Users need to run:
  orion cmd --config examples/my-config.yaml --ack ack/4.22_node-density_ack.yaml

  #### Proposed Behavior

  Orion should automatically detect and load the appropriate ACK file based on:
  1. The OCP version being tested (e.g., 4.22)
  2. The test type being run (e.g., node-density, cluster-density, crd-scale)

  The ACK files already follow a consistent naming convention: {version}_{test-type}_ack.yaml

  #### Implementation Suggestions

  - Parse the version and test type from the config file or test metadata
  - Look for matching ACK file in the ack/ directory
  - Auto-load if found, with an optional --no-ack flag to disable this behavior
  - Log which ACK file was loaded for transparency
  - Continue to allow --ack to override or supplement the auto-loaded ACKs

  ####Benefits

  - Reduces user error from forgetting to specify ACK files
  - Ensures known regressions are always accounted for
  - Simplifies the command line interface
  - Makes ACKs the default rather than opt-in

  #### Files for Reference

  - ACK files are located in ack/ directory
  - Current ACK loading logic: main.py:160-161
  - ACK loader function: orion/config.py (load_ack)

#### Consolidate ACK files
Why consolidate:

  1. Simpler maintenance - One file to update instead of figuring out which of 41 files to edit
  2. No technical risk - The uuid + metric matching is globally unique regardless of file structure
  3. Better auditability - Single source of truth, easy to grep/review
  4. Scales better - Adding new versions doesn't require creating 8+ new files

  Proposed structure:

  ```
ack:
    # === Version 4.22 ===

    # node-density
    - uuid: 679b905a-318c-4b8c-a630-4b2dcc394d25
      metric: ovnCPU_avg
      version: "4.22"        # optional metadata
      test: "node-density"   # optional metadata
      reason: "CPU bumps due to unintentional change in deletion strategy"

    # === Version 4.21 ===

    # cluster-density
    - uuid: 540fb223-bc18-4565-874d-79a5443d7003
      metric: apiserverCPU_avg
      version: "4.21"
      test: "cluster-density"
      reason: "apiserver increase from 4.19 to 4.21..."```

[...]

PR #287, opened 2026-02-18
Automating ack file loading

Balatripura587:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

### Auto-load consolidated ACK file

Orion automatically loads `ack/all_ack.yaml` when it exists, so users no longer need to pass `--ack ack/4.22_node-density_ack.yaml` for the default known regressions. Known regressions are considered by default when running with a config.

### Version and test-type filtering

When the consolidated file is auto-loaded, only ACK entries for the current run are used. Version and test type are taken from the loaded config (first test's metadata: `version_field` / `ocpVersion` and `benchmark.keyword`), with template variables resolved via `--input-vars` (e.g. `version: "{{ version }}"`). Entries in `all_ack.yaml` with matching `version` and `test` are applied; entries without `version`/`test` are still included for backward compatibility.

### `--no-ack` flag

New `--no-ack` flag disables only **automatic** ACK loading (i.e. auto-loading of `ack/all_ack.yaml`). Manually specified files via `--ack` are still loaded and merged when `--no-ack` is set. Log message: `Automatic ACK loading disabled (--no-ack flag)`.

### `--ack` unchanged and additive

The existing `--ack` flag is unchanged. Manually specified ACK files are loaded in addition to the auto-loaded `ack/all_ack.yaml` (when auto-load is not disabled). All ACK sources are merged with deduplication by `(uuid, metric)`.

### Consolidated ACK format

Implementation uses a single consolidated file `ack/all_ack.yaml` with optional per-entry `version` and `test` metadata for filtering. Entries without these fields are treated as applicable to all runs. `auto_detect_ack_file_with_vars()` only looks for `ack/all_ack.yaml`; per-version/test files are deleted

- Related Issue # https://github.com/cloud-bulldozer/orion/issues/261
- Closes # https://github.com/cloud-bulldozer/orion/issues/261


mohit-sheth (approved):
lgtm

PR #291, opened 2026-02-24
Add validation to avoid duplicated metric names

chentex:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

As discussed in chat, there is a problem where if there is a duplicated composed metric name, e.g.: `ovnCPU_avg`

1. There is moment where the process will lose the configuration of the first instance of that duplicated name.
2. Then pandas will resolve this conflict and assign `_x` to one and `_y` to the other
3. And in the end it won't have key values to compare the threshold and the change percentage for any of this metrics, so in case of a Change point in them the whole process will break.

This will check the configuration and build the composed metrics names to validate that there are no duplicates in them, otherwise it will fail, with an error containing the metrics that are in conflict.

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/orion/config.py
+++ b/orion/config.py
@@ -6,6 +6,7 @@ Module file for config reading and loading
 import sys
 import os
 from typing import Any, Dict, List
+from collections import Counter
 import jinja2
 import yaml
 from orion.logger import SingletonLogger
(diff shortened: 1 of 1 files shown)


vishnuchalla (approved):
Changes look good.

PR #323, opened 2026-03-17
Remove value field from agg:

rsevilla87:
## Type of change

- [x] Optimization

## Description

This field is confusing and not actually necessary, it is used to provide a name to the the returned ES bucket metric, however, we can just use metric_of_interest instead with the same result

- Related Issue #
- Closes #

### DIFF
--- a/orion/matcher.py
+++ b/orion/matcher.py
@@ -340,7 +340,7 @@ class Matcher:
             .extra(size=0)
             .sort({timestamp_field: {"order": "desc"}})
         )
-        metric_of_interest = metric["metric_of_interest"]
+        metric_of_interest = metrics["metric_of_interest"]
         agg_type = metrics["agg"]["agg_type"]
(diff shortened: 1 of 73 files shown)


orion/matcher.py
```diff
@@ -362,34 +362,35 @@ def get_agg_metric_query(
...
             percents = metrics["agg"].get("percents", [50, 95, 99])
             uuid_bucket.metric(
-                agg_value, "percentiles",
+                metric_of_interest, "percentiles",
                 field=metrics["metric_of_interest"],
                 percents=percents
             )
         elif agg_type == "count":
             # Count aggregation uses value_count in OpenSearch
-            uuid_bucket.metric(agg_value, "value_count", field=metrics["metric_of_interest"])
```
vishnuchalla, line 358: didn't realize both value and metrics_of_interest have been same for a while. Rebase should prepare it to go.

vishnuchalla (approved):
lgtm

PR #338, opened 2026-04-02
labels.container.keyword filters to CPU metrics

mukrishn:
Add labels.container.keyword filters to CPU metrics in trt-external-payload configs to query specific containers instead of aggregating across all containers in a namespace


examples/trt-external-payload-cluster-density.yaml
```diff
@@ -53,9 +55,23 @@ tests:
         direction: 1
         threshold: 10
 
-      - name: monitoringCPU
+      - name: multusCNIPluginCPU
+        metricName.keyword: containerCPU
+        labels.namespace.keyword: openshift-multus
+        labels.container.keyword: kube-multus-additional-cni-plugins
```
afcollins: Why? Is this a big or little spender?

examples/trt-external-payload-cluster-density.yaml
```diff
@@ -68,6 +84,20 @@ tests:
       - name: ovnCPU
```
mohit-sheth: ovnCPU can be the overall one
and then we need a breakdown by containers

examples/trt-external-payload-cluster-density.yaml
mukrishn: I see, I thought it would be used much. I can remove it then.

examples/trt-external-payload-cluster-density.yaml
mukrishn: I can add all of it by components 👍 and leave this as is.

examples/trt-external-payload-node-density.yaml
```diff
@@ -62,9 +63,75 @@ tests:
...
+        metric_of_interest: value
+        agg:
+          value: cpu
+          agg_type: avg
+        labels:
+          - "[Jira: Networking / ovn-kubernetes]"
+        direction: 1
+        threshold: 10
+
+      - name: ovnCPU-sbdb
```
afcollins: Let's pick a standard name format for these names, especially on net-new.

ovnControllerCPU
ovnkControllerCPU (ok following a pattern)
ovnCPU-northd ( ok, new pattern, adds a dash)
etc.

I think the pattern has already been established [here](https://github.com/mukrishn/orion/blob/2b24acf08c821eda8de6ade1460f52d57153e56d/examples/trt-external-payload-node-density-cni.yaml#L181), let's be consistent.

mukrishn: update: they are being referred in [all_ack.yaml](https://github.com/cloud-bulldozer/orion/blob/main/ack/all_ack.yaml), if we are changing existing config, we should correct ack file as well.

examples/large-scale-cluster-density.yaml
```diff
@@ -99,7 +99,7 @@ tests:
         direction: 1
         threshold: 10
 
-      - name: ovsCPU-irate-all
+      - name: ovsCPUIrateAll
```
afcollins: If we are holistically renaming all the metrics, then I would really like to remove 'Irate' from this name. Reason is: It is more common to use a rate on a cpu metric, and the use of the counter that this was meant to distinguish from is the exception. That one should have 'counter' in the name, or something. This one should really just be 'ovsCPU'.

Maybe we can leave it out of the changes here and have a separate focused change to replace the counter metric with the irate version all at once.

examples/small-scale-node-density-cni.yaml
```diff
@@ -199,7 +199,7 @@ tests:
         direction: 1
         threshold: 10
 
-      - name: ovsMemory-Masters
+      - name: ovsMasterMemory
```
afcollins: If we have an 's' in ovsWorkers above, let's be consistent with 'ovsMasters' here as wel

examples/large-scale-cluster-density.yaml
mukrishn: yes, we can take further changes in a separate PR regards to naming conventions,  this is much required to catch CPU regressions with new container filter.

afcollins: Then I think we should leave it out of this PR to reduce repeated touches to the same lines. Let's omit them from here and can update all later.

mohit-sheth (approved):

afcollins (approved):
LGTM

afcollins:
/lgtm

Failure will happen until branch is rebased, but I ran the configs and they work. Except for crd-scale, which gives the same empty csv outside of this PR.

The values look better. Thank you!

PR #357, opened 2026-04-21
Add yamllint to test requirements

afcollins:
## Type of change

- [x] Bug fix

## Description

yamllint was missing as a dependency, so it would not be found in new installs.

## Testing
1. Create new venv
2. `pip install -r requirements-test.txt`
3. `make lint`


afcollins:
Seems this was missed in the earlier PR that updated lint make target and hasn't surfaced because yamllint was already cached in the runner's pip cache.

paigerube14 (approved):
/lgtm

PR #358, opened 2026-04-21
Standardize metric naming

mukrishn:
Rename hyphenated OVN/OVS metric names to camelCase across all example configs and ack file for consistency and readability.


mukrishn:
@mohit-sheth @afcollins Lets discuss naming conventions here

vishnuchalla:
Commits needs to be signed please.

afcollins (commented):
This is so noisy of a change for now. I have changes coming in mind for queries and config files. Suggest we table this for now until we have a reason to touch all the metrics.

PR #361, opened 2026-04-22
Cast timestamp after proven to be seconds

afcollins:
satisfies pandas to_datetime() FutureWarning on unit

## Type of change

- [x] Bug fix

## Description

This code currently throws a warning when executed. Since we check whether `timestamp` is numeric, we can simply cast it to an int to satisfy the warning.

```
FutureWarning: The behavior of 'to_datetime' with 'unit' when parsing strings is deprecated. In a future version, strings will be parsed as datetime strings, matching the behavior without a '
unit'. To retain the old behavior, explicitly cast ints or floats to numeric type before calling to_datetime.                                                                                                                                                                
    dt = pd.to_datetime(timestamp, unit='s', utc=True)                                                                                                                                                                                                                       
```

## Testing
Unit tests added can be executed via 'pytest'

### DIFF
--- a/orion/tests/test_utils.py
+++ b/orion/tests/test_utils.py
@@ -1,0 +1,87 @@
+"""
+Unit tests for orion/utils.py
+"""
+
+# pylint: disable = redefined-outer-name
+# pylint: disable = missing-function-docstring
+# pylint: disable = import-error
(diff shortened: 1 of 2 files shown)


vishnuchalla (approved):
lgtm

PR #362, opened 2026-04-23
Adds field for pytest to bypass TestResults

afcollins:
## Type of change

- [x] Bug fix

## Description

The class `TestResults` in run_test is not a test class, but pytest thinks it is and throws a warning about it.

This field tells pytest to ignore the confusing name.

- [x] If it is a core feature, I have added thorough tests. (not required)


afcollins:
@chentex maybe we can rename this to not use "Test" for production code during the rewrite/refactoring.

vishnuchalla (approved):
lgtm. Thanks for this change.

PR #366, opened 2026-04-23
Partial revert "Bumping pandas to 2.3.3 and adds build constraint (#355)"

afcollins:
Remove build constraint now that meson 1.11.1 fixed pandas issues

This reverts commit b44f0d971e18d05c5bd8fee116005448e6a609c2.

The build-constraint was only a temporary workaround until meson fixed the inconsistency.

Note: pandas-2.3.3 stays in place


chentex:
/lgtm

Test are failing because we lock the ci-test to run on what it is on main, main still has the constraint line.

afcollins:
Ok. Did that change recently? When I added the build constraint, it updated the pip cache that then benefited all the other PRs before it was merged (i thought)

chentex:
/override "setup-python / setup" "builder" "linter" "tester" "unit_tester"

Issue #360, opened 2026-04-22
[Bug] JSON artifact written before early-changepoint validation, causing inconsistency with viz


mohit-sheth:
html
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/periodic-ci-openshift-eng-ocp-qe-perfscale-ci-main-aws-4.22-nightly-x86-payload-control-plane-6nodes/2046958375308103680/artifacts/payload-control-plane-6nodes/openshift-qe-orion-node-density-cni/artifacts/junit_payload-node-density-cni_viz.html

json
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/periodic-ci-openshift-eng-ocp-qe-perfscale-ci-main-aws-4.22-nightly-x86-payload-control-plane-6nodes/2046958375308103680/artifacts/payload-control-plane-6nodes/openshift-qe-orion-node-density-cni/artifacts/junit_payload-node-density-cni.json

Metric: `ovsCPU-irate-all_avg `

The junit JSON artifact is written before the early-changepoint validation runs, so if a changepoint in the first 5 data points gets cleared by window expansion, the JSON still shows it as a regression while the viz and junit XML correctly show no  changepoint.                                                        
                                                                                                                                                                                                                                                              Move the JSON write at run_test.py:280-284 to after the early-changepoint handling completes (after line 405), so it uses the final validated results.

csheremeta:
@mohit-sheth @vishnuchalla Do we only want to write out the JSON artifact when a changepoint is detected, or do we always want to write out a JSON artifact?

vishnuchalla:
> [@mohit-sheth](https://github.com/mohit-sheth) [@vishnuchalla](https://github.com/vishnuchalla) Do we only want to write out the JSON artifact when a changepoint is detected, or do we always want to write out a JSON artifact?

Its depends on the value specified in `--output-format` [option](https://github.com/cloud-bulldozer/orion/blob/main/main.py#L470), regardless there is a changepoint or not.

vishnuchalla:
@csheremeta if you are trying to correlate this issue description with JSON output artifact, its about correcting that JSON artifact content to reflect what it needs to be in terms of changepoints rather than deciding when to write it out.

csheremeta:
[...]

PR #367, opened 2026-04-24
fix outdated json after extended window issue

ArthurChenCoding:
## Type of change
- [x] Bug fix

## Description

The JSON is created before the window extension and not updated after the window extension, more description on the issue below.

## Fix

Following the suggestion from the issue, I delayed the JSON creation after the potential window extension.

- Related Issue # https://github.com/cloud-bulldozer/orion/issues/360

- [x] I have performed a self-review of my code, unit test, and e2e local test

## Testing
Added a unit test to ensure the JSON is updated in case of a window extension.
Performed e2e local testing with 4.22.0-0.nightly-2026-04-10-145151 (the same version mentioned in the issue) and confirm the HTML viz file and JSON are now aligned.


ArthurChenCoding:
@Balatripura587 is this issue mentioned in the PR message related to your prior window work?

ArthurChenCoding:
I do not think I can do a smaller fix besides some syntactic sugar. 

The diff seems huge, but the "if test_flag:" loop at line 423 is almost the same, just unindented.

orion/run_test.py
```diff
@@ -405,53 +414,59 @@ def analyze(test, kwargs, is_pull = False) -> AnalyzeResult:
                         display_fields=kwargs.get("display"),
                     )
```
vishnuchalla, line 407: Why can't the new logic be re-written in this block instead of completely removing it?

orion/run_test.py
```diff
@@ -278,18 +277,25 @@ def analyze(test, kwargs, is_pull = False) -> AnalyzeResult:
     expanded_algorithm = None
     # This is env is only present in prow ci
     prow_job_id = os.getenv("PROW_JOB_ID")
-    if kwargs["output_format"] != cnsts.JSON and prow_job_id and prow_job_id.strip():
-        testname, result_data, _ = algorithm.output(cnsts.JSON)
-        output_file_name = f"{os.path.splitext(kwargs['save_output_path'])[0]}_{testname}.json"
-        with open(output_file_name, 'w', encoding="utf-8") as file:
-            file.write(str(result_data))
+    write_sidecar_json = (
```
vishnuchalla: How is this block different from the above block at Line #280 which is removed?

orion/tests/test_run_test.py
```diff
@@ -0,0 +1,265 @@
+"""
```
vishnuchalla: If we want to add tests, lets take it in a separate PR. Lets keep the changes minimal here, IMHO.

vishnuchalla (changes requested):
Why do we need all these additional changes while the issue description clearly highlights

> Move the JSON write at run_test.py:280-284 to after the early-changepoint handling completes (after line 405), so it uses the final validated results.

ArthurChenCoding:
Thanks, refinements are in place. Although the prior approach also handles the non-regression path via deferred write, I forgot the fact that we only need to ensure an updated JSON when regression exists

Balatripura587 (commented):
/lgtm

Balatripura587 (approved):

vishnuchalla (approved):
Thanks, /lgtm

PR #371, opened 2026-04-27
Fix dockerfile and github action

chentex:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Update dockerfile and image.
Makes github action work only when a new tag is created.

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/.github/workflows/image-push.yaml
+++ b/.github/workflows/image-push.yaml
@@ -2,31 +2,44 @@ name: Image Push
 on:
   workflow_dispatch:
   push:
-    branches:
-      - master
-      - main
+    tags:
+      - 'v*'
 
 jobs:
   build_image:
     name: Image Push
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/image-push.yaml
```diff
@@ -2,31 +2,44 @@ name: Image Push
 on:
   workflow_dispatch:
   push:
-    branches:
-      - master
-      - main
+    tags:
```
vishnuchalla, line 5: Do we want to make it only when release is triggered? By any idea on where we are using orion images?

chentex: My idea is to only have release image, I could add a develop tag, for main.
As far as I know no one is using images

vishnuchalla: Sound good.
/lgtm

vishnuchalla (approved):
/approve

PR #374, opened 2026-04-28
Display kube-burner-ocp version

mmnabeel317:
## Summary

Add the **kube-burner-ocp version** to all hover tooltips in the interactive HTML visualizations, so users can see which version of kube-burner-ocp was used for each run without leaving the chart.

The version is fetched from `jobSummary` documents in the benchmark Elasticsearch index (where kube-burner-ocp already writes it as the `version` field), mapped per UUID, and displayed in the hover alongside existing fields like OCP version, UUID, and build URL.

### `orion/matcher.py`
- Added `get_kb_version()` method to the `Matcher` class
- Queries `jobSummary` docs from the benchmark index using `_source` filtering (only fetches `uuid` and `version` fields) for minimal overhead
- Returns a `{uuid: version}` dict, stripping the `@commitHash` suffix for clean display
- Deduplicates on UUID and excludes `garbage-collection` jobs

### `orion/utils.py`
- Calls `match.get_kb_version()` in `process_test()` after the index switches to the benchmark index
- Merges the result as a `kbVersion` column into `merged_df`, following the same pattern used for `buildUrl` and `ocpVersion`

### `orion/visualization.py`
- Extracts `kbVersion` from the dataframe with a graceful fallback to `"N/A"`
- Added `kube-burner-ocp: <version>` line to all three hover templates:
  - Data-point hover (main time-series markers)
  - Changepoint hover (diamond markers)
  - ACK marker hover (acknowledged changepoints)

### `orion/tests/test_visualization.py`
- Added `kbVersion` column to the `sample_dataframe` fixture
- Added assertion that the generated HTML contains `kube-burner-ocp` in hover text

## Design decisions

- **No external repo changes required** — the `version` field already exists in ES from kube-burner-ocp
- **Lightweight** — `_source` filtering keeps the ES query small (2 fields per doc)
- **Graceful fallback** — non-kube-burner workloads (ingress-perf, netperf) show `N/A`
- **Historical data works** — the `version` field has been in `JobSummary` since early kube-burner versions

## Reference 
- **PJ-Rehearse Job**: https://github.com/openshift/release/pull/78442

/cc @mohit-sheth

### DIFF
--- a/orion/matcher.py
+++ b/orion/matcher.py
@@ -227,6 +227,45 @@ class Matcher:
         runs = [hit.to_dict()["_source"] for hit in all_hits]
         return runs
 
+    def get_kb_version(self, uuids: List[str],
+                       timestamp_field: str = "timestamp") -> Dict[str, str]:
+        """Get kube-burner-ocp version for each UUID from jobSummary docs.
+
+        Args:
+            uuids (list): list of uuids
+            timestamp_field (str): timestamp field in data
+
+        Returns:
+            dict: mapping of uuid to kube-burner-ocp version string
+        """
+        query = Q(
+            "bool",
+            filter=[
+                Q("terms", **{self.uuid_field + ".keyword": uuids}),
+                Q("match", metricName="jobSummary"),
+                ~Q("match", **{"jobConfig.name": "garbage-collection"}),
+            ],
+        )
+        search = (
+            Search(using=self.es, index=self.index)
+            .query(query)
+            .source([self.uuid_field, "version"])
+            .extra(size=self.search_size)
+            .sort({timestamp_field: {"order": "desc"}})
+        )
+        try:
+            all_hits = self.query_index(search, return_all=True)
+        except Exception:  # pylint: disable=broad-except
+            return {}
+        kb_versions = {}
+        for hit in all_hits:
+            src = hit.to_dict()["_source"]
+            uid = src.get(self.uuid_field)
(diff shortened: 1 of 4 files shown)


orion/matcher.py
```diff
@@ -227,6 +227,45 @@ def match_kube_burner(self, uuids: List[str],
         runs = [hit.to_dict()["_source"] for hit in all_hits]
         return runs
 
+    def get_kb_version(self, uuids: List[str],
```
vishnuchalla, line 230: @mmnabeel317 have you considered the data that `--github-context` flag emits to pull out kube-burner version?

mmnabeel317: I didn't use that since the github_context data (produced by --github-repos) only gets populated for changepoints, and it returns a list of GitHub releases and commits that happened between the previous and current changepoint timestamps.Just to be sure this is the snippet I'm referring to.

```
#algorithm.py L67-113 — github_context is only set inside the changepoint loop
for key, value in change_points_by_metric.items():
    for change_point in value:
        ...
        github_client = self._get_github_client()
        if github_client and not dataframe_json[index].get("github_context"):
            context = github_client.get_change_context(...)
```

chentex:
I think this should be the correct approach to this #383

chentex:
Closing this since #383 
Review documentation here https://github.com/cloud-bulldozer/orion/blob/main/docs/configuration.md#metadata-metrics

PR #380, opened 2026-04-30
adding tests for PR 367

ArthurChenCoding:
## Type of change

Adding test

## Description

Adds unit tests for the behaviour fixed in [PR 367](https://github.com/cloud-bulldozer/orion/pull/367), which moved and corrected
The Prow CI JSON sidecar file-writing logic inside `analyze()` in `run_test.py`.

Related Issue # https://github.com/cloud-bulldozer/orion/issues/360

- [x] I have performed a self-review of my code
image

## what is tested
| Test | Behaviour verified |
|---|---|
| `test_file_written_with_proper_json_format` | Sidecar file is created with valid, indented JSON (via `json.dumps`) when `output_format != json`, `PROW_JOB_ID` is set, and a regression is detected |
| `test_file_not_written_when_output_format_is_json` | No sidecar file is written when the requested output format is already JSON |
| `test_file_not_written_without_prow_job_id` | No sidecar file is written when `PROW_JOB_ID` env var is absent |
| `test_file_not_written_when_prow_job_id_is_blank` | A whitespace-only `PROW_JOB_ID` is treated as absent — no sidecar written |
| `test_file_not_written_when_no_regression` | No sidecar file is written when the algorithm reports no regression (`test_flag=False`) |
| `test_file_content_reflects_post_expansion_data` | Sidecar file content reflects the **final** `result_data_json` after window expansion, not the pre-expansion snapshot — the core guarantee of #367 |

### DIFF
--- a/orion/tests/test_run_test.py
+++ b/orion/tests/test_run_test.py
@@ -1,0 +1,236 @@
+"""
+Unit tests for orion/run_test.py
+
+Focused on the Prow CI JSON file-writing behaviour fixed in PR #367:
+  - file is written *after* window expansion using json.dumps (not str())
+  - file is only written when output_format != JSON and PROW_JOB_ID is present
+  - file is not written when there is no regression
+"""
+
+# pylint: disable = redefined-outer-name
+# pylint: disable = missing-function-docstring
+# pylint: disable = import-error
+# pylint: disable = missing-class-docstring
+
+import json
+import logging
+import os
+from unittest.mock import MagicMock, patch
+
+import pandas as pd
+import pytest
+
+from orion.logger import SingletonLogger
+import orion.constants as cnsts
+from orion.run_test import analyze
+
+
+# ---------------------------------------------------------------------------
+# Fixtures / helpers
+# ---------------------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


ArthurChenCoding:
@Balatripura587 here is the unit tests addition for the prior window extension bug fix, can you take a look?

chentex:
/lgtm

Can we make sure we have tests on the PR with the code, it is difficult to follow having just the unit tests

ArthurChenCoding:
@chentex Thanks for the feedback, by " tests on the PR with the code" you mean integration test? can you point me an example for me to mimic and implement?

afcollins (approved):

afcollins:
> @chentex Thanks for the feedback, by " tests on the PR with the code" you mean integration test? can you point me an example for me to mimic and implement?

I think what he means is to submit the production code and unit tests in the same PR in the future. I agree, where possible.

If we are adding tests for code that currently doesn't have any coverage, that would be an exception.

Otherwise, net-new code and changes to existing code *should* have unit tests alongside it.

ArthurChenCoding:
thanks for explaining.

PR #381, opened 2026-05-01
Attach HTML visualization files to JIRA issues on auto-create

mohit-sheth:
When --jira-auto-create and --viz are both enabled, the generated HTML visualization file is now attached to the corresponding JIRA issue. This lets reviewers download the interactive chart directly from the ticket. https://redhat.atlassian.net/browse/PERFSCALE-4731 

- `create_ack()` returns issue key (Optional[str]) instead of bool
- ` --jira-auto-create` works independently of` --jira-ack` (helps with testing already ack'd uuids)
- Extract` _attach_viz_to_jira()` helper for clean separation
- Demote noisy `aggregated_metric_data` log to DEBUG

### DIFF
--- a/main.py
+++ b/main.py
@@ -161,7 +161,7 @@ def format_jira_description(regression: dict, metric_name: str, pct_change: floa
     return desc
 
 
-def auto_create_jira_issues(regression_data: list, provider: AckProvider, logger) -> int:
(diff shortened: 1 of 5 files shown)


mukrishn (approved):
/lgtm

PR #385, opened 2026-05-05
Adds ES query batching to save time on Orion execution

chentex:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds ES query batching to save time on Orion execution

- Related Issue #
- Closes #

## Testing

## Summary by CodeRabbit

* **New Features**
  * Single-request batched retrieval for aggregated and standard metrics, with configurable UUID field and percentile/count support.

* **Refactor**
  * Metric processing reorganized to prefer chunked batch dispatch, preserve/restore per-metric metadata, and centralize aggregation-to-table construction.

* **Bug Fixes**
  * Percentile aggregations now require an explicit percents list; missing percents are reported and skipped. Config validation enforces this.

* **Tests**
  * Expanded coverage for batching, chunking, fallback behavior, percentiles, counts, and config validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cloud-bulldozer/orion/pull/385)

### DIFF
--- a/orion/config.py
+++ b/orion/config.py
@@ -85,6 +85,13 @@ def load_config(config_path: str, input_vars: Dict[str, Any]) -> Dict[str, Any]:
         for metric in test["metrics"]:
             metric_name = test["name"] + ":" + metric["name"]
             if "agg" in metric:
+                if metric["agg"]["agg_type"] == "percentiles" and "percents" not in metric["agg"]:
+                    logger.error(
+                        "Metric '%s' in test '%s' has agg_type 'percentiles' "
+                        "but no 'percents' list defined in the agg block",
+                        metric["name"], test["name"]
+                    )
+                    sys.exit(1)
                 metric_name = f"{metric_name}:{metric['agg']['agg_type']}"
             elif "metric_of_interest" in metric:
                 metric_name = f"{metric_name}:{metric['metric_of_interest']}"
--- a/orion/matcher.py
+++ b/orion/matcher.py
@@ -520,6 +520,9 @@ class Matcher:
                 field = metric["metric_of_interest"]
                 filtered = uuid_bucket[agg_name]
 
+                if filtered.doc_count == 0:
+                    continue
+
                 row = {
                     self.uuid_field: uuid_val,
                     timestamp_field: ts_val,
@@ -601,7 +604,10 @@ class Matcher:
         results = {m["name"]: [] for m in metrics_list}
         for doc in runs:
             for metric_name, match_fields in filter_fields_by_metric:
-                if all(doc.get(k) == v for k, v in match_fields.items()):
+                if all(
+                    doc.get(k.replace(".keyword", "")) == v
+                    for k, v in match_fields.items()
+                ):
                     results[metric_name].append(doc)
                     break
 
--- a/orion/matcher.py
+++ b/orion/matcher.py
@@ -578,7 +578,8 @@ class Matcher:
                 for k, v in metric.get("not", {}).items()
             ]
             should_clauses.append(Q("bool", must=filter_clauses + not_clauses))
-            filter_fields_by_metric.append((metric["name"], match_fields))
+            not_fields = metric.get("not", {})
+            filter_fields_by_metric.append((metric["name"], match_fields, not_fields))
 
         query = Q(
             "bool",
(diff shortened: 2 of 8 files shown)


orion/matcher.py
```diff
@@ -413,6 +413,212 @@ def parse_agg_results(
...
+            ]
+            not_clauses = [
+                ~Q("match", **{k: v})
+                for k, v in metric.get("not", {}).items()
+            ]
+            metric_filter = Q("bool", must=metric_filter_clauses + not_clauses)
+            filtered_bucket = uuid_bucket.bucket(agg_name, "filter", metric_filter)
+
+            if agg_type == "percentiles":
+                percents = metric["agg"].get("percents", [50, 95, 99])
```
afcollins: These are hard-coded and not taken from the config?

chentex: Thanks for the catch @afcollins

chentex: That is the default is no percentiles are given in the config, which should be an error so I'll validate that beforehand and add a failure here since it should not had happened.

orion/config.py
```diff
@@ -85,6 +85,13 @@ def load_config(config_path: str, input_vars: Dict[str, Any]) -> Dict[str, Any]:
         for metric in test["metrics"]:
             metric_name = test["name"] + ":" + metric["name"]
             if "agg" in metric:
+                if metric["agg"]["agg_type"] == "percentiles" and "percents" not in metric["agg"]:
+                    logger.error(
+                        "Metric '%s' in test '%s' has agg_type 'percentiles' "
+                        "but no 'percents' list defined in the agg block",
+                        metric["name"], test["name"]
+                    )
+                    sys.exit(1)
```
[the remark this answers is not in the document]
chentex: Ignore, known behavior

orion/matcher.py
```diff
@@ -413,6 +418,218 @@ def parse_agg_results(
...
+        Returns:
+            Dict mapping metric name -> list of hit _source dicts.
+        """
+        if not metrics_list:
+            return {}
+
+        excluded_keys = {"name", "metric_of_interest", "not"}
+        filter_fields_by_metric = []
+        should_clauses = []
+        for metric in metrics_list:
```
afcollins, line 581: Can we parametrize how many metrics to batch into a single request?

A config with 56 metrics can overwhelm the request, with an error like the following, while a config with 21 metrics runs fine. We might protect against that by fixing the number of metrics we lump into a query batch.

```
File "pandas/_libs/join.pyx", line 188, in pandas._libs.join.full_outer_join
numpy._core._exceptions._ArrayMemoryError: Unable to allocate 93.7 TiB for an array with shape (12878280455040,) and data type int64
```

chentex: Thanks for this test @afcollins

chentex: I've pushed a patch to try to fetch 15 metrics at a time.

orion/utils.py
```diff
@@ -56,44 +57,194 @@ def get_metric_data(
...
+
+            meta_by_name[metric["name"]] = {
+                "labels": labels, "direction": direction, "threshold": threshold,
+                "correlation": correlation, "context": context, "timestamp": ts,
+            }
+
+            if "agg" in metric:
+                agg_metrics.append(metric)
+            else:
+                std_metrics.append(metric)
```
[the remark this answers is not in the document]
chentex: no, they get reattached again alter

orion/utils.py
```diff
@@ -58,48 +59,206 @@ def get_metric_data(
...
+
+    def _build_agg_dataframe(self, data, metric, match, timestamp_field="timestamp"):
+        """Build a DataFrame from pre-fetched aggregation data."""
+        aggregation_value = metric["metric_of_interest"]
+        aggregation_type = metric["agg"]["agg_type"]
+
+        if aggregation_type == "percentiles":
+            percentile_prefix = f"{aggregation_value}_{aggregation_type}_"
+            agg_columns = [k for k in data[0].keys()
+                           if k.startswith(percentile_prefix)] if data else []
```
afcollins, line 228: Just a nitpick, but this is getting kinda ugly to have spread around. I feel bad for introducing it :/

Maybe we can refactor to a pattern that makes it less special or at least something re-usable so we aren't looping through each time.

orion/utils.py
```diff
@@ -58,48 +59,206 @@ def get_metric_data(
...
+                            meta["timestamp"],
+                        )
+                        dataframe_list.append(metric_df)
+                        self._restore_meta(metric, meta)
+                        self._register_columns([single_name], metric, meta, metrics_config, metadata_columns)
+                    except Exception as e2:
+                        self.logger.error("Couldn't get metric %s: %s", name, e2)
+                        self._restore_meta(metric, meta)
+
+    def _build_agg_dataframe(self, data, metric, match, timestamp_field="timestamp"):
```
afcollins, line 220: If this method is analogous to `process_standard_metric`, could it read to be more similar? method name, argument order (data vs. preloaded_data), code is formatted differently because of variable name length.

If this method name is more desirable going forward, then we could plan to migrate legacy code for consistency.

afcollins (approved):
I think it looks good.

Some comments I made were thoughts I had while reviewing.

Thanks for the optimization!

orion/utils.py
chentex: Thanks for your comments @afcollins .
@coderabbitai create an issue with this comment.

orion/utils.py
chentex: Thanks I will take a look at this
@coderabbitai create an issue with this comment.

PR #398, opened 2026-05-20
Revert "adding tests for PR 367"

afcollins:
Unit tests added are failing.

Reverts cloud-bulldozer/orion#380


afcollins:
I don't see that tests were ever run in the previous PR.

chentex:
Thank you for this @afcollins
I think for security we only run what it is on main branch, but I think the unit tests should also run from the PR branch I will check if we need to add some changes to that. this could have been avoided.

ArthurChenCoding:
Thank you both @afcollins @chentex for reverting this. I can see that PR 400 and 404 addresses and split the CI so that unit test runs on the branch, which is safer.

PR #399, opened 2026-05-20
Add pylint make target

afcollins:
... and use it from github workflow.

Also, omit tests from pylint.

## Type of change

- [x] Refactor

## Description

This changes makes it easier for contributors to lint locally instead of relying on github workflow.

Also, removes tests from pylint, where we might tend to do things that won't satisfy linters, that we add pylint hints to ignore anyway. :)


afcollins:
cc: @chentex

chentex:
/skip CI tests\ / unit_tester / test

PR #402, opened 2026-05-25
Change default direction from 0 to 1 to detect only increases

Sandeepyadav93:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [X] Optimization
- [ ] Documentation Update

## Description

Most metrics (CPU, latency, memory, etc.) are regressions when they increase, so direction: 1 (positive changes only) is the more sensible default. Users who want to detect both increases and decreases can explicitly set direction: 0, and those tracking metrics where decreases are bad (like throughput) can use direction: -1.

Updated documentation to reflect the new default and clarify when direction: 0 is needed for bidirectional change detection.

- Related Issue #
- Closes #

- [X] I have performed a self-review of my code.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -273,8 +273,8 @@ UUID field ensures that a set of test results from the same job are grouped toge
 ### Direction
 Controls which types of changes to detect:
 
-- `direction: 1` - Show only positive changes (increases)
-- `direction: 0` - Show both positive and negative changes (default)
+- `direction: 1` - Show only positive changes (increases) **(default)**
+- `direction: 0` - Show both positive and negative changes
 - `direction: -1` - Show only negative changes (decreases)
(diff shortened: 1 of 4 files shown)


rsevilla87:
Can you sign the commit please @Sandeepyadav93 ?

rsevilla87 (approved):
/lgtm
The change looks safe to me, in the metrics where direction is not currently specified, the direction is positive, so this change should help to prevent observing false positives in case any metric or KPI improves.

chentex:
Thank you @Sandeepyadav93 
@coderabbitai Open a new issue tittle "Refactor VIZ code", description "It seems we are looping through change point in different places of the code, and simple changes like setting the default for the direction need to be done in all those places."

PR #411, opened 2026-06-10
Consolidate config files

rsevilla87:
## Type of change

- [x] Refactor

## Description

Create generic configuration files for node-density, node-density-cni, cd-v2, node-density-pods-udn and crd-scale, the metadata fingerprint will be generated automatically by the `ocp-metadata` binary, and it won't be needed to have the same configuration files for the different node scales:

```shell
$ ./bin/ocp-metadata  | jq .
{
  "distribution": "openshift",
  "microshift": false,
  "platform": "External",
  "clusterType": "self-managed",
  "ocpVersion": "4.20.4",
  "ocpMajorVersion": "4.20",
  "k8sVersion": "v1.33.5",
  "masterNodesType": "BM.Optimized3.36",
  "masterNodesCount": 3,
  "workerNodesCount": 3,
  "totalNodes": 3,
  "sdnType": "OVNKubernetes",
  "clusterName": "ci-dg8ms",
  "fips": false,
  "publish": "External",
  "workerArch": "amd64",
  "controlPlaneArch": "amd64",
  "ipsec": false,
  "ipsecMode": "Disabled"
}
```

Changes in the prow side:
- https://github.com/openshift/release/pull/80284

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/examples/cluster-density.yaml
+++ b/examples/cluster-density.yaml
@@ -9,7 +9,7 @@ tests:
       workerNodesCount: {{ workerNodesCount }}
       benchmark.keyword: cluster-density-v2
       wildcard:
-        ocpVersion: "{{ version }}*"
+        ocpVersion: "{{ ocpMajorVersion }}*"
       networkType: OVNKubernetes
       jobType: "{{ jobtype | default('periodic') }}"
       pullNumber: "{{ pull_number | default(0) }}"
--- a/examples/cluster-density.yaml
+++ b/examples/cluster-density.yaml
@@ -5,7 +5,7 @@ tests:
       clusterType: {{ clusterType }}
       masterNodesType: {{ masterNodesType }}
       masterNodesCount: {{ masterNodesCount }}
-      workerNodesType: {{ worker_nodes_type }}
+      workerNodesType: {{ workerNodesType }}
       workerNodesCount: {{ workerNodesCount }}
       benchmark.keyword: cluster-density-v2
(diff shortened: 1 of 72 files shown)


examples/cluster-density.yaml
```diff
@@ -0,0 +1,141 @@
+tests:
+  - name: cluster-density-v2
+    metadata:
+      platform: {{ platform }}
```
afcollins: Please quote jinja substitution for formatter

examples/cluster-density.yaml
```diff
@@ -0,0 +1,141 @@
...
+      networkType: OVNKubernetes
+      jobType: "{{ jobtype | default('periodic') }}"
+      pullNumber: "{{ pull_number | default(0) }}"
+      organization: "{{ organization | default('') }}"
+      repository: "{{ repository | default('') }}"
+      fips: "{{ fips | default('false') }}"
+      not:
+        stream: okd
+
+    metrics:
```
afcollins, line 22: We should start using the metrics inheritance for any new files we add

rsevilla87: We better address that in a follow-up PR, to prevent adding massive changes. the intention is to remove the old config files once we merge this one.

examples/cluster-density.yaml
rsevilla87: Fixed

PR #412, opened 2026-06-11
Add jobName.keyword wildcard filter to scale config metadata

ArthurChenCoding:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [X] Optimization
- [ ] Documentation Update

## Description

Enables filtering ES queries by job name in all scale-based configs. Defaults to '*' (match all) for backward compatibility. When the jobname env var is set, only runs from matching CI jobs are compared, preventing data mixing across job variants (e.g., control-plane vs etcdencrypt vs loaded-upgrade).

- Related Issue #
- Closes #

- [X] I have performed a self-review of my code.

## Testing
I am merging this to test this PR https://github.com/openshift/release/pull/79477

### DIFF
--- a/examples/label-small-scale-cluster-density.yaml
+++ b/examples/label-small-scale-cluster-density.yaml
@@ -10,6 +10,7 @@ tests:
       fips: "{{ fips | default('false') }}"
       wildcard:
         ocpVersion: "{{ version }}*"
+        jobName.keyword: "{{ jobname | default('*') }}"
       not:
         stream: okd
         # networkType: OVNKubernetes
--- a/examples/large-scale-cluster-density.yaml
+++ b/examples/large-scale-cluster-density.yaml
@@ -9,6 +9,7 @@ tests:
       benchmark.keyword: cluster-density-v2
       wildcard:
         ocpVersion: "{{ version }}*"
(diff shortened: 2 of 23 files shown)


examples/label-small-scale-cluster-density.yaml
```diff
@@ -10,6 +10,7 @@ tests:
       fips: "{{ fips | default('false') }}"
       wildcard:
         ocpVersion: "{{ version }}*"
+        jobName.keyword: "{{ jobname | default('*') }}"
```
afcollins, line 13: Mixing `.keyword` and wildcard `*` is an antipattern.

If we want fuzzy matching, we can remove `.keyword`.

Also, How do we also prevent that `node-density` doesn't also match on `node-density-cni`?

chentex: I agree this does not meet the criteria to be a `wildcard`.

afcollins (changes requested):
wildcard usage with `.keyword` exact match is an anti pattern. Please prefer fuzzing match or some other way.

afcollins:
Replaced by #460

PR #414, opened 2026-06-25
fix: support nested field traversal in get_results_batch metric routing

eifrach:
## Type of change

- [ ] Refactor
- [ ] New feature
- [X] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

  Problem

  When metric configs reference nested OpenSearch fields using dot-notation (e.g. tags.namespace, tags.pod), get_results_batch silently drops all matching documents.

  The batch query correctly retrieves documents from OpenSearch using dot-notation field paths (e.g. Q("match", **{"tags.namespace": "openshift-apiserver"})), but the subsequent Python-side routing step uses doc.get("tags.namespace") to assign each document to the right metric. Since OpenSearch returns documents with nested structure
  ({"tags": {"namespace": "..."}, ...}), a flat dict.get() with a dotted key always returns None, so no document ever matches any metric and all metrics silently return empty data.

  Indexes with flat top-level fields (e.g. resource_type, component) are unaffected — this only manifests when metric filter fields are nested under an object like tags.

  Fix

  Added a _get_nested() static helper that traverses nested dicts using dot-notation keys, and replaced the two doc.get() calls in the Python-side routing loop with it. For non-nested fields, the helper falls back to plain doc.get(), so existing tests are unaffected.

  Impact

  Without this fix, any metric config that filters on nested fields (e.g. tags.namespace: openshift-apiserver) produces a dataframe with all-NaN values, and EDivisive finds no changepoints — resulting in 0 reported failures even when clear regressions exist in the data.

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/orion/matcher.py
+++ b/orion/matcher.py
@@ -618,11 +618,11 @@ class Matcher:
             matched = False
             for metric_name, match_fields, not_fields in filter_fields_by_metric:
                 matches_positive = all(
-                    doc.get(k.replace(".keyword", "")) == v
+                    self._get_nested(doc, k.replace(".keyword", "")) == v
                     for k, v in match_fields.items()
                 )
                 matches_negative = all(
-                    doc.get(k.replace(".keyword", "")) != v
+                    self._get_nested(doc, k.replace(".keyword", "")) != v
                     for k, v in not_fields.items()
                 )
                 if matches_positive and matches_negative:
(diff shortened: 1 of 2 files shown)


chentex:
Thank you for the contribution and the bug fix!
Can you add a unit test scenario for this @eifrach

eifrach:
> Thank you for the contribution and the bug fix! Can you add a unit test scenario for this @eifrach

Will do, thanks

PR #416, opened 2026-06-26
Remove IgnoreGlobal and IgnoreGlobalMetrics parameters

rsevilla87:
This pull request removes support for the `IgnoreGlobal` and `IgnoreGlobalMetrics` test-level options, which previously allowed tests to opt out of inheriting global config and metrics. Documentation, test cases, and configuration examples related to these options have all been deleted, and the code that handled them in `orion/config.py` has been simplified accordingly.

These parameters are not very useful, if the user just doesn't want to inherit parts of the configuration it should not use neither `metricsFile` nor  `parentConfig`

## Summary by CodeRabbit

* **Bug Fixes**
  * Configuration inheritance now consistently includes parent metadata and metrics when present.
  * Removed special-case behavior that let individual tests skip inherited global settings.
* **Documentation**
  * Removed outdated documentation for ignoring inherited global metadata and metrics.
* **Tests**
  * Updated inheritance-related test coverage to reflect the new merged configuration behavior.


chentex:
@vishnuchalla PTAL since this was a feature you specifically asked for.

PR #420, opened 2026-06-26
Add release workflow action

chentex:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds release binary workflow

- Related Issue #
- Closes #

## Testing

## Summary by CodeRabbit

* **New Features**
  * Added automated release publishing for tagged builds.
  * Release artifacts now include a Linux AMD64 binary attached to GitHub Releases.


.github/workflows/release.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - '*'
+
+permissions:
+  contents: write
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
```
rsevilla87: Why are you using the commit hash rather than the tag?

```suggestion
      - uses: actions/checkout@v6
```

btw, v7 is already around

.github/workflows/release.yaml
```diff
@@ -0,0 +1,51 @@
...
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Set up Python
+        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6
```
rsevilla87, line 21: Same as above

.github/workflows/release.yaml
```diff
@@ -0,0 +1,51 @@
...
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          pip install pyinstaller
+          pip install .
+
+      - name: Bake version
+        run: echo "__version__ = \"${GITHUB_REF#refs/tags/}\"" > version.py
+
+      - name: Build standalone binary
```
rsevilla87, line 36: Have you considered publishing the package to pypi?

chentex: It crossed my mind, but I don't know if we have to have any consideration when doing that as Red Hat?

.github/workflows/release.yaml
chentex: coderabbits comment about security concern, sha is immutable as v7 is not.

.github/workflows/release.yaml
chentex: The other thing I am thinking now is this might be the same issue that it would need to download the dependencies and that is an issue we have sometimes when github fails to pull otava or such.

rsevilla87: yeah, that's correct, pypi is just a distribution method, it doesn't create a self contained module

PR #423, opened 2026-07-01
Remove stale and unused configuration files

rsevilla87:
Remove several stale and unused configuration files, I haven't found any reference to them in prow, I wonder if they are still useful in other areas

/cc @afcollins @chentex @jtaleric @mohit-sheth 

## Summary by CodeRabbit

* **Chores**
  * Removed a large set of example benchmark configurations across small, medium, and large scale scenarios.
  * Cleaned up several cluster-density, node-density, payload, and UDN example files by deleting outdated test definitions and metric checks.
  * Also removed readout examples for control-plane node-density variants.

* **Documentation**
  * Example files now contain less sample content, reducing duplication and simplifying the available configuration set.


chentex:
Merging this will break the following integration tests
```⏺ Yes, 6 of the 26 files being removed in PR #423 are directly referenced in test.bats:

  ┌────────────────────────────────────────────────────┬──────────────────────────────────────────┐
  │                    Removed file                    │             Test(s) using it             │
  ├────────────────────────────────────────────────────┼──────────────────────────────────────────┤
  │ examples/label-small-scale-cluster-density.yaml    │ Lines 165, 169, 182, 186 (4 tests)       │
  ├────────────────────────────────────────────────────┼──────────────────────────────────────────┤
  │ examples/small-scale-cluster-density.yaml          │ Line 210 (anomaly detection test)        │
  ├────────────────────────────────────────────────────┼──────────────────────────────────────────┤
  │ examples/small-scale-node-density-cni.yaml         │ Line 214 (anomaly detection with window) │
  ├────────────────────────────────────────────────────┼──────────────────────────────────────────┤
  │ examples/small-scale-node-density.yaml             │ Referenced by name match                 │
  ├────────────────────────────────────────────────────┼──────────────────────────────────────────┤
  │ examples/trt-external-payload-cluster-density.yaml │ Lines 218, 222, 226 (3 tests)            │
  ├────────────────────────────────────────────────────┼──────────────────────────────────────────┤
  │ examples/trt-external-payload-node-density.yaml    │ Lines 230, 234, 238, 242, 246 (5 tests)  │
  └────────────────────────────────────────────────────┴──────────────────────────────────────────┘
  ```

chentex (changes requested):
Please add the changes to the bats tests before we merge this

examples/trt-external-payload-cluster-density-percentiles.yaml
afcollins, line 1: I added this one to purely be an example for how to use percentiles.

afcollins:
What is the motivation? If we want to distinguish those we actually use as production configs, is it time to move those select few out? Then examples can be a collection of true examples?

rsevilla87:
> What is the motivation? If we want to distinguish those we actually use as production configs, is it time to move those select few out? Then examples can be a collection of true examples?

because they are not in use? the small/medium/large-scale config file approach is not necessary anymore, so I don't find a good reason to keep and maintain these files
agree, we should move the production ones to a another dir

examples/trt-external-payload-cluster-density-percentiles.yaml
rsevilla87: restored

afcollins:
> > What is the motivation? If we want to distinguish those we actually use as production configs, is it time to move those select few out? Then examples can be a collection of true examples?
> 
> because they are not in use? the small/medium/large-scale config file approach is not necessary anymore, so I don't find a good reason to keep and maintain these files agree, we should move the production ones to a another dir

Ok, sounds good.

/lgtm

jtaleric:
For the time being @rsevilla87 lets leave the TRT configs for orion-mcp (legacy) we are working to close that gap.

Balatripura587:
Can we merge this after we refactor orion-mcp to align with new configs?

PR #430, opened 2026-07-13
Add Orion configs for 5 CUDN density test jobs

mmnabeel317:
## Summary

Adds Orion regression detection configs for 5 CUDN density test jobs (24-node AWS):

| CI Job | Config | Differentiator |
|--------|--------|----------------|
| cudn-density-single-ns-250-24nodes | `small-scale-cudn-density-single-ns-250.yaml` | Renamed from `l2-single-ns` |
| cudn-pod-churn-250-24nodes | `small-scale-cudn-pod-churn-250.yaml` | Pod churn (30m, 50%) |
| cudn-churn-250-24nodes | `small-scale-cudn-churn-250.yaml` | Namespace churn (3 cycles, 10m) |
| cudn-density-single-ns-500-24nodes | `small-scale-cudn-density-single-ns-500.yaml` | 500 iterations, m6a.4xlarge |
| cudn-incremental-700-24nodes | `small-scale-cudn-incremental-700.yaml` | Incremental 50/step, linear |

### Changes

- **Shared metrics file** — `metrics/cudn-density-metrics.yaml` with 17 OVN/OVS metrics (8 CPU + 8 memory + podReadyLatency P99). All 5 configs inherit via `metricsFile`.
- **kbCfg metadata** — Each config uses `kbCfg.*` fields from real ES data (`iterations`, `churnMode`, `churnDuration`, `incrementalPattern`, etc.) for precise job filtering.
- **incremental-700 override** — Overrides `podReadyLatency` with `agg_type: max` to collapse 14 incremental steps into one value per run (worst-case P99 at full 700-CUDN scale).
- **No jobName.keyword in metrics** — Removed from all configs; kube-burner uses different job names across churn/incremental modes and versions.

### Verified outputs

All 5 configs tested against live ES — no cross-contamination between job types:
https://gist.github.com/mmnabeel317/f9fe43c1cfde974ef7e9c09bf1cb9091

/cc @mohit-sheth


examples/small-scale-cudn-density-l2-single-ns-500.yaml
```diff
@@ -0,0 +1,238 @@
...
+    metadata:
+      platform: AWS
+      masterNodesType: m6a.4xlarge
+      masterNodesCount: 3
+      workerNodesType: m6a.4xlarge
+      workerNodesCount: 24
+      benchmark.keyword: cudn-density
+      kbCfg.iterations: "500"
+      kbCfg.namespacesPerCudn: "1"
+      kbCfg.churnDuration: "0s"
```
mohit-sheth: does this config work without this?
adding churn config would be confusing I think

mmnabeel317: Yes , tested it now, Its redundant , the namespacePerCudn also seems redundant as its set to 1 in every config, will remove that as well.

mohit-sheth (approved):
lgtm

afcollins:
These files are all pretty similar. Can we instead use inheritance for the metrics?

examples/small-scale-cudn-churn-250.yaml
```diff
@@ -0,0 +1,222 @@
...
+      workerNodesType: m6a.2xlarge
+      workerNodesCount: 24
+      benchmark.keyword: cudn-density
+      kbCfg.iterations: "250"
+      kbCfg.churnMode: "namespaces"
+      kbCfg.churnCycles: "3"
+      kbCfg.churnDuration: "10m0s"
+      wildcard:
+        ocpVersion: "{{ version }}*"
+        upstreamJob.keyword: "*cudn-churn-250*"
```
afcollins: Wildcard usage of exact match provided by '.keyword' is an anti-pattern.

If wildcard is the right way, then let's remove '.keyword'.

afcollins: If I'm not wrong, it should suffice to pull `upstreamJob: "cudn-churn-250"` outside of the wildcard block (without `.keyword`), and that should match. Please try this too.

afcollins:
Let's also amend this commit to have an extra newline between the first line and the rest of the commit message.

Without the newline, it is being renderd as a single line in the `git show-branch` output:
```
! [main] Add release workflow action (#420)
 * [add-cudn-orion-configs] Add Orion configs for 5 CUDN density test jobs - small-scale-cudn-density-single-ns-250: renamed from l2-single-ns, matches cudn-density-single-ns-250-24nodes - small-scale-cudn-pod-churn-250: new config for cudn-pod-churn-250-24nodes with kbCfg filtering - small-scale-cudn-churn-250: new config for cudn-churn-250-24nodes (namespace churn mode) with kbCfg filtering - small-scale-cudn-density-l2-single-ns-500: new config for cudn-density-single-ns-500-24nodes with kbCfg filtering - small-scale-cudn-incremental-700: new config for cudn-incremental-700-24nodes with kbCfg filtering and max aggregation for podReadyLatency Added kbCfg metadata fields from real ES data for precise run matching.
--
 * [add-cudn-orion-configs] Add Orion configs for 5 CUDN density test jobs - small-scale-cudn-density-single-ns-250: renamed from l2-single-ns, matches cudn-density-single-ns-250-24nodes - small-scale-cudn-pod-churn-250: new config for cudn-pod-churn-250-24nodes with kbCfg filtering - small-scale-cudn-churn-250: new config for cudn-churn-250-24nodes (namespace churn mode) with kbCfg filtering - small-scale-cudn-density-l2-single-ns-500: new config for cudn-density-single-ns-500-24nodes with kbCfg filtering - small-scale-cudn-incremental-700: new config for cudn-incremental-700-24nodes with kbCfg filtering and max aggregation for podReadyLatency Added kbCfg metadata fields from real ES data for precise run matching.
 * [add-cudn-orion-configs^] fix issue where numpy version mismatch with otava (#429)
 * [add-cudn-orion-configs~2] fix: support nested field traversal in get_results_batch metric routing (#414)
 * [add-cudn-orion-configs~3] Move to use pypi otava (#428)
 * [add-cudn-orion-configs~4] Revert release artifact, not working as expected (#427)
+* [main] Add release workflow action (#420)
```

mmnabeel317:
> These files are all pretty similar. Can we instead use inheritance for the metrics?

You're right @afcollins , I have Created examples/metrics/cudn-density-metrics.yaml with the shared 17-metric block (8 OVN/OVS CPU + 8 memory + podReadyLatency P99). All 5 configs now use metricsFile: metrics/cudn-density-metrics.yaml and are ~20 lines each.

The incremental-700 config overrides podReadyLatency inline with agg_type: max to collapse 14 incremental steps into a single value per run. The shared file keeps podReadyLatency as a standard metric so it stays generic for other use cases.

examples/small-scale-cudn-churn-250.yaml
mmnabeel317: Tested this. Using upstreamJob: cudn-churn-250 as a plain match field, does not work correctly. Orion builds ES match queries which use the default OR operator. So cudn-density-single-ns-250 tokenizes to cudn OR density OR single OR ns OR 250, and the query matches any document containing any of those tokens. This causes cross-contamination: the single-ns-250 config was pulling in pod-churn and churn runs because they all share cudn and 250 tokens.

The kbCfg fields help for churn/incremental (they have unique values like churnMode: namespaces), but single-ns-250 and single-ns-500 don't have enough unique kbCfg to prevent false matches on their own.

Kept wildcard: upstreamJob.keyword since it is the only way to do substring matching on the full Prow job name without modifying Orion's query builder. Open to alternative suggestions if there is a better pattern.

jtaleric (commented):
Generally I am good with these changes - I am curious about the `.keyword` discussion though. 

Nice work @mmnabeel317

examples/small-scale-cudn-churn-250.yaml
afcollins: Sounds good to me.

Thank you for testing out the alternatives. I appreciate the extra effort and context.

I am curious how wildcard.upstreamJob (without .keyword) would work? #431 already merged. :/

mmnabeel317:
Hey @afcollins , Due to the changes in #431 , I removed `upstreamJob.keyword` and switched to using `kbCfg.*` only. Please take a look at the changes and let me know what you think.

examples/small-scale-cudn-density-single-ns-250.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - name: aws-small-scale-cudn-density-single-ns-250
+    metadata:
+      platform: AWS
+      masterNodesType: m6a.2xlarge
+      masterNodesCount: 3
+      workerNodesType: m6a.2xlarge
+      workerNodesCount: 24
+      benchmark.keyword: cudn-density
+      kbCfg.iterations: "250"
+      kbCfg.churnDuration: "0s"
```
afcollins, line 12: From the name 'single-ns', should this also have: `kbCfg.namespacesPerCudn: "1"` that I see in the next file?

edit: I see the [conversation](https://github.com/cloud-bulldozer/orion/pull/430#discussion_r3571153498) about removing it. I was wondering whether this is a unique enough fingerprint to distinguish from other runs.

mmnabeel317: The kbCfg.namespacesPerCudn: "1" is the same across all 5 CUDN jobs so it doesn't help with differentiation here. For single-ns-250, kbCfg.churnDuration: "0s" is what separates it from pod-churn-250 ("30m0s") and churn-250 ("10m0s"). The single-ns-500 config does include namespacesPerCudn: "1" because that's what distinguishes it from the multi-ns-500 job.

mmnabeel317: https://gist.github.com/mmnabeel317/c46594bd2ecb2d48e1f6e5f764f587c8

Here is the updated gist that shows no cross contamination for reference

afcollins, line 12:
```suggestion
      upstreamJob.keyword: "periodic-ci-openshift-eng-ocp-qe-perfscale-ci-main-aws-{{ version }}-nightly-x86-cudn-density-single-ns-250-24nodes"
```

afcollins: Is this what we want to achieve? To filter on a particular job as a stream?

I tested it with version 4.22 and 5.0 and it works as it looks.

I don't see any other variables that we need wildcard matching for, and it's more direct than trying to detect fingerprint through kbCfg.

Let me know what you think.

afcollins: Thanks for the gist and explanation. I see that this uniquely identifies a job today, but I just worry that it leaves room to pull in other configs unknowingly.

I see that all versions are included in the same output (like output_table_aws-small-scale-cudn-density-single-ns-250.txt), is that the intended use of this config?

Also, see the other suggestion.

mohit-sheth: I prefer the kbCfg approach, it doesnt couple orion to prow configs and is less of a maintenance burden
I understand that `upstreamJob.keyword` is cleaner and the chances of pulling wrong runs is lower..
 but we can enhance kbCfg to match even more flags if required

mmnabeel317: That's not the intended use, but I didn't specify the version under the --input-vars flag, so it fetches all the runs for config validation.

mmnabeel317: I Agree with Mohit on kbCfg decreasing maintenance burden. Also worth noting that `benchmark.keyword: cudn-density` is already scoping all configs to only CUDN runs, so the kbCfg fields are differentiating within an already narrow set, which lowers the risk of pulling wrong runs.

afcollins: Sure, sounds good.

afcollins: Ok! That makes sense the version was left out. Sounds good to me.

afcollins (approved):
lgtm

PR #443, opened 2026-07-21
Add static and dynamic metric expansion

### LINKED ISSUE #425: [RFE] Dynamic bucket discovery
**Is your feature request related to a problem? Please describe.**

When dealing with highly dimensional metrics that contain multiple labels—such as containerCPU, which has unique labels for each container and pod—configuring Orion becomes incredibly tedious. Currently, users must manually define a separate metric configuration block for every single label combination they want to ingest or monitor.

```yaml
- name: ovnCPU-ovncontroller
  metricName.keyword: containerCPU
  labels.namespace.keyword: openshift-ovn-kubernetes
  labels.container.keyword: ovn-controller
  metric_of_interest: value
  agg:
    agg_type: avg
  direction: 1
  threshold: 10

- name: ovnCPU-northd
  metricName.keyword: containerCPU
  labels.namespace.keyword: openshift-ovn-kubernetes
  labels.container.keyword: northd
  metric_of_interest: value
  agg:
    agg_type: avg
  direction: 1
  threshold: 10
# Repeat for every other container...
```

**Describe the solution you'd like**

Introduce a `groupBy`  parameter that grants Orion the ability to dynamically discover and aggregate these labels. This would allow users to consolidate repetitive definitions into a single, clean block.

With this feature, the previous example could be drastically reduced to:

```yaml
- name: ovnCPU
  metricName.keyword: containerCPU
  labels.namespace.keyword: openshift-ovn-kubernetes
  groupBy: 
  - labels.container.keyword
  metric_of_interest: value
  agg:
    agg_type: avg
  direction: 1
  threshold: 10
```

For instance, the single ovnCPU block above should automatically expand or append the grouped label value to generate results for:

- ovnCPU-northd
- ovnCPU-ovn-controller
- ovnCPU-nbdb
- ovnCPU-sbdb
- ovnCPU-ovnkube-controller
- etc

afcollins:
## Type of change
- [x] New feature
## Description
Allow metric definitions to use a `fan_out` key that expands a single template into multiple metrics with ${var} placeholder substitution and field overrides, reducing config duplication.

Add group_by support for dynamic metric expansion

Query OpenSearch for distinct label values scoped by metric filters and UUIDs, then expand metric templates per discovered value. Enables auto-discovery instead of listing values manually in fan_out entries.

Document fan_out and group_by metric expansion features

Add Metric Expansion section to configuration.md covering static fan_out and dynamic group_by with examples, comparison table, and usage guidance.

Assisted-by: Claude Code
- Closes #425
## Testing
Run against 4.22 with 5d lookback:
```VERSION=4.22 orion --column-group-size=4 --collapse --node-count true --config examples/trt-external-payload-cluster-density-group-by.yaml --lookback 5d --hunter-analyze```
See output in this [gist](https://gist.github.com/afcollins/b4cfc9efefc8ac177147da9392c877fa).

## Summary by CodeRabbit

* **New Features**
  * Added metric expansion using `fan_out`, allowing one template to generate multiple metrics with variable substitution and per-entry overrides.
  * Added dynamic `group_by` expansion, creating metrics for distinct values discovered from OpenSearch data.
  * Added documentation and configuration examples for metric expansion.
  * Expanded support across filters, labels, aggregations, correlation, and metadata metrics.

* **Tests**
  * Added comprehensive coverage for static and dynamic metric expansion, substitutions, overrides, and edge cases.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -540,6 +540,104 @@ not:
     - "sidecar"
 ```
 
+## Metric Expansion
+
+Orion provides two ways to create multiple metrics from a single template, reducing duplication in configuration files.
+
+### `fan_out` — Static Expansion
+
+Use `fan_out` when you know the exact values you want to expand. Each entry in the `fan_out` list produces one metric, with `${var}` placeholders substituted and matching fields overridden.
+
+```yaml
+metrics:
+  - name: "${label}CPU"
+    metricName.keyword: containerCPU
+    labels.namespace.keyword: "${namespace}"
+    metric_of_interest: value
+    agg:
+      agg_type: avg
+    labels:
+      - "[Jira: ${jira}]"
+    direction: 1
+    threshold: 10
+    fan_out:
+      - label: apiserver
+        namespace: openshift-kube-apiserver
+        jira: kube-apiserver
+      - label: multus
+        namespace: openshift-multus
+        jira: multus
+      - label: ovn
+        namespace: openshift-ovn-kubernetes
+        jira: ovn-kubernetes
+```
+
+This expands into three metrics: `apiserverCPU`, `multusCPU`, and `ovnCPU`, each with the correct namespace filter and JIRA label. Without `fan_out`, you would need to write each metric definition separately.
+
+**How it works:**
+
+1. Each `fan_out` entry is a dict of variable names to values
+2. All `${var}` placeholders in string values are replaced recursively (including nested dicts and lists)
+3. If a `fan_out` entry key matches an existing metric field, the entry value overrides it directly
+4. Each expanded metric is a deep copy — mutations to one do not affect others
+5. The `fan_out` key is removed after expansion
+
+**Field override example:**
+
+```yaml
+- name: "ovnCPU-${container}"
+  metricName.keyword: containerCPU
+  labels.container.keyword: "${container}"
+  fan_out:
+    - container: northd
+    - container: ovncontroller
+      labels.container.keyword: ovn-controller  # overrides the substituted value
+```
+
+The second entry uses the override value `ovn-controller` instead of the substituted value `ovncontroller`.
+
+### `group_by` — Dynamic Expansion
+
+Use `group_by` when you don't know the values ahead of time and want Orion to discover them from OpenSearch. Orion queries for all distinct values of the specified field, scoped by the metric's other filters and the test's UUIDs, then creates one metric per discovered value.
+
+```yaml
+metrics:
+  - name: "${labels.namespace.keyword}CPU"
+    metricName.keyword: containerCPU
+    metric_of_interest: value
+    agg:
+      agg_type: avg
+    direction: 1
+    threshold: 10
+    group_by:
+      - labels.namespace.keyword
+```
+
+If the OpenSearch data contains `containerCPU` documents with namespaces `openshift-etcd`, `openshift-kube-apiserver`, and `openshift-multus`, this expands into three metrics — one per namespace — with `labels.namespace.keyword` set as both a filter and a `${labels.namespace.keyword}` substitution variable.
+
+**How it works:**
+
+1. `group_by` accepts a list with exactly one field name (multi-field support is planned)
+2. Orion runs a terms aggregation on that field, filtered by the metric's other fields (e.g., `metricName.keyword`) and the test's UUIDs
+3. For each discovered value, a deep copy of the template is created with:
+   - The `group_by` field set as a filter (e.g., `labels.namespace.keyword: openshift-etcd`)
+   - All `${field.name}` placeholders substituted with the discovered value
+4. If no values are found, the metric is skipped with a warning
+5. The `group_by` key is removed after expansion
+
+**When to use `group_by` vs `fan_out`:**
+
+| | `fan_out` | `group_by` |
+|---|---|---|
+| Values known ahead of time | Yes | No — discovered at runtime |
+| Supports multiple variables per entry | Yes | No — one field per `group_by` |
+| Requires OpenSearch connection | No — expands at config load | Yes — queries during test processing |
+| Use case | Fixed set of components | Discover all namespaces/containers/labels |
+
+**Combining with other features:**
+
+`group_by` works with all other metric features — aggregations, correlation, labels, `not` filters, and metadata metrics. The expanded metrics are identical to hand-written metrics after expansion.
+
 ## Test-Level Settings
 
 Settings that can be applied at the test level and inherited by all metrics:
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -582,20 +582,6 @@ This expands into three metrics: `apiserverCPU`, `multusCPU`, and `ovnCPU`, each
 4. Each expanded metric is a deep copy — mutations to one do not affect others
 5. The `fan_out` key is removed after expansion
 
-**Field override example:**
-
-```yaml
-- name: "ovnCPU-${container}"
-  metricName.keyword: containerCPU
-  labels.container.keyword: "${container}"
-  fan_out:
-    - container: northd
-    - container: ovncontroller
-      labels.container.keyword: ovn-controller  # overrides the substituted value
-```
-
-The second entry uses the override value `ovn-controller` instead of the substituted value `ovncontroller`.
-
 ### `group_by` — Dynamic Expansion
 
 Use `group_by` when you don't know the values ahead of time and want Orion to discover them from OpenSearch. Orion queries for all distinct values of the specified field, scoped by the metric's other filters and the test's UUIDs, then creates one metric per discovered value.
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by afcollins** [comment]
A great way to shine light on all the data hiding in the shadows.

It even pointed out two new potential regressions:


```
Regression(s) found :                                                                                                                                                                                                                                                        
--------------------------------------------------                                                                                                                                                                                                                           
Test: payload-cluster-density-v2:                                                                                                                                                                                                                                            
Changepoint at:      4.22.0-0.nightly-2026-07-20-143422                                                                                                                                                                                                                      
Previous version:    4.22.0-0.nightly-2026-07-20-035241                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                             
Affected Metrics                                                                                                                                                                                                                                                             
+-------------------------------------------------+----------+---------------------+----------+                                                                                                                                                                              
| Metric                                          |    Value | Percentage change   | Labels   |                                       
+=================================================+==========+=====================+==========+                                       
| containerCPU-machine-healthcheck-controller_avg | 0.100265 | 14.03%              | []       |                                                                                                                                                                              
+-------------------------------------------------+----------+---------------------+----------+                                       
                                                                                                                                                                                                                                                                             
Other PRs in payload (1):                                                                                                                                                                                                                                                    
  - https://github.com/openshift/cluster-capi-operator/pull/624                                                                                                                                                                                                              
--------------------------------------------------                                                                                                                                                                                                                           
Test: payload-cluster-density-v2:                                                                                                                                                                                                                                            
Changepoint at:      4.22.0-0.nightly-2026-07-18-004714                                                                                                                                                                                                                      
Previous version:    4.22.0-0.nightly-2026-07-17-163757                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                             
Affected Metrics                                                                                                                                                                                                                                                             
+---------------------------------+---------+---------------------+----------+                                                                                                                                                                                               
| Metric                          |   Value | Percentage change   | Labels   |                                                                                                                                                                                               
+=================================+=========+=====================+==========+                                                                                                                                                                                               
| containerCPU-northd_avg         | 2.49724 | 20.65%              | []       |                                                                                                                                                                                               
| containerCPU-ovn-controller_avg | 1.49171 | 29.52%              | []       |                                                                                                                                                                                               
+---------------------------------+---------+---------------------+----------+                                                                                                                                                                                               
                                                                                                                                                                                                                                                                             
Other PRs in payload (4):                                                                                                                                                                                                                                                    
  - https://github.com/openshift/installer/pull/10685                                                                                                                                                                                                                        
  - https://github.com/openshift/console/pull/16768                                       
  - https://github.com/openshift/api/pull/2930                                                                                                                                       
  - https://github.com/openshift/ovn-kubernetes/pull/3304                                                                                                                            
```

Batching is also works great here:
```
2026-07-21 13:55:35,549 - Orion      - INFO - file: utils.py - line: 139 - Batching aggregation metrics chunk 181-195 of 364
```

**Comment by afcollins** [comment]
cc @chentex @rsevilla87 @mohit-sheth

**Comment by chentex** [comment]
Hey @afcollins this is looking great, can you run it with a debug and let me see the query that is formed, you can paste it in the same gist.

I think we might have a situation with this that for some strange reason a metric only exists on one of the UUIDs and not on the rest of them, this would generate a table with one value and tons of NaN or 0, how should we manage those metrics?

I am on board with this looks clean, works and gives us more visibility into our metrics.

**Comment by afcollins** [comment]
Updated the gist with the full log (sensitive bits redacted) and queries extracted to JSONs.

>I think we might have a situation with this that for some strange reason a metric only exists on one of the UUIDs and not on the rest of them, this would generate a table with one value and tons of NaN or 0, how should we manage those metrics?

I think that's valid. If a subset of runs have data, we should still return it to the user. I don't think erroring out until all fields match across all runs makes sense. What's another alternative?

Looking at the queries, I see:

queries.63.json: Find UUIDs against perf_scale_ci
queries.64.json: Query perf_scale_ci again with UUIDs to resolve ocpVersion
queries.{65..68}.json: (four queries) Resolve dynamic fields for each metricName and field. `must:[ { terms: uuid.keyword: [...] } , {match: metricName.keyword} ], aggs: { group_values: { terms: field:`
queries.{69..93}.json: batches to query all columns for all runs. Snippet:
```
  "aggs": {
    "uuid": {
      "terms": { "field": "uuid.keyword", "size": 11 },
      "aggs": {
        "time": { "avg": { "field": "timestamp" }
        },
        "containerCPU-alertmanager": {
          "filter": {
            "bool": {
              "must": [
                { "match": { "metricName.keyword": "containerCPU" } },
                { "match": { "labels.container.keyword": "alertmanager" } }
              ]
            }
          },
          "aggs": { "value": { "avg": { "field": "value" } }
          }
        },
```

**Comment by chentex** [comment]
Thank you @afcollins for updating, it was way more than i asked, but useful, I like what I am seeing in the queries. I am surprised the batching works so well, even with so many metrics discovered.

> What's another alternative?

The only other alternative I see is adding an exists clause for all UUIDs but I think that would over complicate the process, I am happy with what you have now.

If you take care of the coderabbit comment and everyone is OK I would go ahead and merge this.

File: orion/utils.py [hunk 1]
```diff
@@ -576,7 +578,11 @@ def process_test(
             options["baseline"],
             options["node_count"],
         )
-        # get metrics data and dataframe
+        # expand groupBy metric templates using live OpenSearch data
+        test["metrics"] = expand_group_by(
+            test["metrics"], match, uuids, self.logger
+        )
+
```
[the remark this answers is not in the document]
afcollins: I agree, it would be nice. However, your proposed fix to except Exception fails pylint `W0718: Catching too general exception Exception (broad-exception-caught)` . What exceptions specifically would you recommend handling?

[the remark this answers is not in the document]
afcollins: What do you think about handling OpenSearchException instead? Or in addition to ConnectionError?

afcollins:
Updated with exception handling and new test. Please review

orion/utils.py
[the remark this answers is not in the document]
afcollins: Fixed. Added to matcher.py. Please check there and close this if sufficient.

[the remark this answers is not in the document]
afcollins: updated now. please check.

orion/matcher.py
```diff
@@ -705,3 +707,68 @@ def dotDictFind(self, data: dict, find: str) -> str:
...
+        ]
+
+        query = Q("bool", must=must_clauses, must_not=not_clauses)
+
+        search = (
+            Search(using=self.es, index=self.index)
+            .query(query)
+            .extra(size=0)
+        )
+        search.aggs.bucket("group_values", "terms", field=field, size=1000)
```
vishnuchalla, line 755: Acknowledging that `1000` is big enough for us to get distinct bucket sizes.

vishnuchalla:
I like this approach, simplifies querying and configs.

docs/configuration.md
```diff
@@ -540,6 +540,104 @@ not:
...
+
+**How it works:**
+
+1. Each `fan_out` entry is a dict of variable names to values
+2. All `${var}` placeholders in string values are replaced recursively (including nested dicts and lists)
+3. If a `fan_out` entry key matches an existing metric field, the entry value overrides it directly
+4. Each expanded metric is a deep copy — mutations to one do not affect others
+5. The `fan_out` key is removed after expansion
+
+**Field override example:**
```
rsevilla87: What's the use case of the field override feature @afcollins ?

orion/matcher.py
rsevilla87: I find difficult to see a metric with more than 1000 different label values :)

rsevilla87: Well, a `pod` or `node` label might be the case, but it doesn't make much sense to use them

docs/configuration.md
afcollins: `fan_out` overall is a more consolidated representation of the multi-line definition for each test.

It's an extension of templating for static expansion, instead of dynamic expansion.

One use case is if we have a handful of values we want to repeat (OVN-k containers, for example) that we know ahead of time, we could set them in this more condensed format.

Another could be setting specific threshold for other label combinations.

Admittedly, it was a v1 and unrelated to the dynamic expansion. So if we don't see a use, then we can leave it out for now.

orion/matcher.py
afcollins: AFAIK, 'container' is the largest cardinality and only <150

docs/configuration.md
afcollins: I removed that bit from the docs. Did you mean something else?

rsevilla87 (commented):
/lgtm
/approve

docs/configuration.md
rsevilla87: Awesome, thanks!

PR #450, opened 2026-07-29
Confidence indicators

chentex:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Every detected changepoint is automatically annotated with a statistical confidence indicator based on:
- **Cohen's d effect size** — magnitude of the shift relative to data variability (primary indicator)
- **Welch's t-test p-value** — included as descriptive context only; it is **not** calibrated for post-selection inference since the changepoint was already selected from this same data
### Labels
Labels are driven by Cohen's d thresholds (Cohen 1988), with p-value shown inline for reference:
| Label | Meaning |
|-------|---------|
| `Large shift (d=1.20, p=0.001)` | d >= 0.8 — strong evidence of a meaningful metric shift |
| `Moderate shift (d=0.60, p=0.03)` | 0.5 <= d < 0.8 — moderate-magnitude shift |
| `Small shift (d=0.30, p=0.02)` | 0.2 <= d < 0.5 — small but detectable effect |
| `Negligible shift (d=0.10, p=0.85)` | d < 0.2 — negligible practical impact |
| `Degenerate variance — shift detected but effect size undefined` | Both segments have zero variance with different means |
| `Insufficient data` | Fewer than 2 data points on either side of the changepoint |
| `Anomaly detection — shift confidence not applicable` | IsolationForest detects outliers, not sustained shifts |
### Segment boundaries
For E-Divisive, before/after segments are bounded by **raw detector boundaries** (all changepoints found by Otava before Orion applies direction/threshold/ACK/correlation filtering). This prevents a filtered recovery changepoint from dilating the comparison win
dow and masking a genuine earlier shift.
For CMR, confidence is computed against the **original (uncollapsed) dataframe**, preserving the baseline standard deviation across all historical runs rather than collapsing to a single averaged row.
### Per-metric `is_changepoint`
Each metric object in JSON output now includes an explicit `is_changepoint` boolean field, replacing the previous inference from `percentage_change != 0`.
### Additional changes
- `ci_95` field is always present in JSON output (set to `null` when confidence cannot be computed)
- SciPy added as a runtime dependency (`pyproject.toml`, `requirements.txt`, `setup.py`)
- `test.bats`: removed tests duplicated with `test-local.bats`
- `test-local.bats`: assertions updated to match new label format and percent formatting
- `test_matcher_batch.py`: added `pylint: disable=protected-access` suppression for `TestGetNested`
- Full documentation in `docs/usage.md`

- Related Issue #
- Closes #

## Testing

## Summary by CodeRabbit

- **New Features**
  - Added per-changepoint confidence indicators for detected changepoints (label, statistical significance, effect size, sample sufficiency, and sample counts).
  - Confidence is now shown in text summaries, standalone reports, and serialized JSON, and included in generated JUnit results.

- **Documentation**
  - Documented the confidence categories, thresholds, and how “Insufficient data” is determined, including algorithm-specific before/after splitting.

- **Bug Fixes / Output Improvements**
  - Improved changepoint highlighting logic and standardized JUnit percent formatting.

- **Chores**
  - Added SciPy as a runtime dependency.

### Text
```
Regression(s) found :
--------------------------------------------------
Test: udn-density-pods:
Changepoint at:      5.0.0-0.nightly-2026-08-14-034145
Previous version:    5.0.0-0.nightly-2026-08-13-145619

Affected Metrics
+-----------------+----------+------------+---------------------------------+-----------------------------------------+
| Metric          |    Value | % Change   | Confidence                      | Labels                                  |
+=================+==========+============+=================================+=========================================+
| ovnCPU-sbdb_avg | 0.792783 | 22.01%     | Large shift (d=3.88, p=7.8e-05) | ['[Jira: Networking / ovn-kubernetes]'] |
+-----------------+----------+------------+---------------------------------+-----------------------------------------+
```
### JSON
```
      "ovnCPU-sbdb_avg": {
        "value": 0.7927828041,
        "percentage_change": 22.006700959322767,
        "is_changepoint": true,
        "labels": [
          "[Jira: Networking / ovn-kubernetes]"
        ],
        "confidence": {
          "p_value": 7.767061601173493e-05,
          "cohens_d": 3.881568911408574,
          "label": "Large shift (d=3.88, p=7.8e-05)",
          "sufficient_data": true,
          "sample_size_before": 37,
          "sample_size_after": 6,
          "mean_before": 0.6904412324213095,
          "mean_after": 0.8423845697401298,
          "std_before": 0.038994230333000604,
          "std_after": 0.04021246475662423,
          "ci_95": [
            0.10977912382081302,
            0.1941075508168275
          ]
        }
      },
```
### JUNIT
```
		<testcase name="[Jira: Networking / ovn-kubernetes] ovnCPU-sbdb_avg regression detection" timestamp="1787224031">
			<failure>
+----+--------------------------------------+----------------------+-------------------+------------------+---------------------+---------------------------------+
|    | uuid                                 | timestamp            |   ovnCPU-sbdb_avg | is_changepoint   | percentage_change   | Confidence                      |
|----+--------------------------------------+----------------------+-------------------+------------------+---------------------+---------------------------------|
|  0 | 7101150f-56f8-4a43-891d-56957c659b59 | 2026-07-31T11:50:24Z |          0.715141 | False            | 0.00%               |                                 |
...                              |
| 36 | 1bd7a246-5e73-46fa-9e9e-1ce755ff6a0e | 2026-08-13T17:28:48Z |          0.763616 | False            | 0.00%               |                                 |
| 37 | c288d7bc-d5d5-42ba-952f-e4edaca5298c | 2026-08-14T06:32:55Z |          0.792783 | True             | 22.01%              | Large shift (d=3.88, p=7.8e-05) | -- changepoint
...
| 42 | afda75e1-78a9-423b-b6c5-8857b2058d62 | 2026-08-20T05:25:43Z |          0.859382 | False            | 0.00%               |                                 |
+----+--------------------------------------+----------------------+-------------------+------------------+---------------------+---------------------------------+
         </failure>
		</testcase>
```


chentex:
@coderabbitai please review again

chentex:
@coderabbitai review

vishnuchalla:
I like this change, can we have a flag to output the change points only after a certain confidence score? That way integration in CI will be simple.

jtaleric:
> I like this change, can we have a flag to output the change points only after a certain confidence score? That way integration in CI will be simple.

I think regardless of the confidence score we let CI stay with all found regressions - i think the additional context will help Chaibot make better decisions long term.

vishnuchalla:
> > I like this change, can we have a flag to output the change points only after a certain confidence score? That way integration in CI will be simple.
> 
> I think regardless of the confidence score we let CI stay with all found regressions - i think the additional context will help Chaibot make better decisions long term.

Correct, but more precise we are the better it is. From the token usage reduction standpoint 💸

jtaleric:
> > > I like this change, can we have a flag to output the change points only after a certain confidence score? That way integration in CI will be simple.
> > 
> > 
> > I think regardless of the confidence score we let CI stay with all found regressions - i think the additional context will help Chaibot make better decisions long term.
> 
> Correct, but more precise we are the better it is. From the token usage reduction standpoint 💸

yeah - that could help reduce the token aspect -- we could pass that via orion-mcp

jtaleric:
@chentex this looks great! One aspect I think we can change (not here, but on a follow-on) is the term `labels` - maybe we just update the JIRA Labels to call it specific to JIRA so we don't overload a single term.  We will have Confidence labels and JIRA lables... 🤢

orion/confidence.py
```diff
@@ -0,0 +1,167 @@
...
+            sufficient_data=False,
+            sample_size_before=n_before,
+            sample_size_after=n_after,
+            mean_before=float(np.mean(before)) if n_before > 0 else None,
+            mean_after=float(np.mean(after)) if n_after > 0 else None,
+            std_before=float(np.std(before, ddof=1)) if n_before > 1 else None,
+            std_after=float(np.std(after, ddof=1)) if n_after > 1 else None,
+        )
+
+    _, p_value = stats.ttest_ind(before, after, equal_var=False)
```
sjug: The changepoint was selected from these same values, so testing it again with an ordinary Welch test makes random noise look significant too often. The reported ci_95 has the same problem. Please calibrate by resampling the full detector-and-selection pipeline or by using separate confirmation data. Otava's existing window statistics may be reusable descriptively, but only if their selection guarantees are understood.

orion/confidence.py
```diff
@@ -0,0 +1,167 @@
...
+    if cohens_d >= 0.2:
+        return f"Possible [{d_str}] (small shift [{p_str}])"
+    return f"Statistically significant [{d_str}] but trivial [{p_str}]"
+
+
+def _get_segments(algorithm_name, data, changepoint_index):
+    """Split data into before/after segments based on algorithm type."""
+    if algorithm_name == cnsts.CMR:
+        return data[:-1], data[-1:]
+    return data[:changepoint_index], data[changepoint_index:]
```
sjug: This replaces the detector's local window with the entire prefix and suffix. A later recovery or reversal can therefore cancel a genuine earlier shift and label it "Noise." Please compare segments bounded by neighboring changepoints, or reuse the detector's window statistics.

orion/confidence.py
```diff
@@ -0,0 +1,167 @@
...
+    if cohens_d >= 0.8:
+        return f"Likely real [{d_str}] (large shift [{p_str}])"
+    if cohens_d >= 0.5:
+        return f"Likely real [{d_str}] (moderate shift [{p_str}])"
+    if cohens_d >= 0.2:
+        return f"Possible [{d_str}] (small shift [{p_str}])"
+    return f"Statistically significant [{d_str}] but trivial [{p_str}]"
+
+
+def _get_segments(algorithm_name, data, changepoint_index):
```
sjug: IsolationForest identifies unusual individual observations, not necessarily a lasting before/after shift. Treating every anomaly as a permanent split can dilute a real outlier and label it "Noise." Please use detector-specific confidence logic or clearly describe this as persistence testing rather than anomaly confidence.

orion/confidence.py
```diff
@@ -0,0 +1,167 @@
...
+    std_before = np.std(before, ddof=1)
+    std_after = np.std(after, ddof=1)
+    pooled_std = math.sqrt(
+        ((n_before - 1) * std_before ** 2 + (n_after - 1) * std_after ** 2)
+        / (n_before + n_after - 2)
+    )
+
+    if pooled_std == 0:
+        cohens_d = (
+            float("inf") if mean_before != mean_after else 0.0
```
sjug: This becomes Infinity in JSON, which strict parsers such as JavaScript's JSON.parse reject. The statistical comparison is undefined when both groups have zero variance, so please return nullable statistics with a reason such as degenerate_variance. Checking this before calling SciPy would also avoid the precision-loss warnings.

orion/confidence.py
```diff
@@ -0,0 +1,167 @@
...
+            "label": self.confidence_label,
+            "sufficient_data": self.sufficient_data,
+            "sample_size_before": self.sample_size_before,
+            "sample_size_after": self.sample_size_after,
+            "mean_before": self.mean_before,
+            "mean_after": self.mean_after,
+            "std_before": self.std_before,
+            "std_after": self.std_after,
+        }
+        if self.ci_95 is not None:
```
sjug: The documentation says ci_95 is null when unavailable, but this omits the field. Please always include `"ci_95": null` so consumers do not need to handle two different object shapes.

orion/confidence.py
```diff
@@ -0,0 +1,167 @@
...
+            result["ci_95"] = list(self.ci_95)
+        return result
+
+
+def _map_label(p_value, cohens_d):
+    """Map p-value and Cohen's d to a human-readable confidence label."""
+    d_str = f"{cohens_d:.2f}" if not math.isinf(cohens_d) else "inf"
+    p_str = f"{p_value:.2f}"
```
sjug: Small p-values are rounded to 0.00, and a high p-value does not mean the shift itself is trivial. Please label the values and use wording such as `Not statistically significant [d=1.50, p=0.30]`. Formatting p-values with `:.2g` would also preserve values such as 5e-10.

orion/utils.py
```diff
@@ -848,8 +848,14 @@ def create_record(record):
             ),
             metric_name: record["metrics"][metric_name]["value"],
             "is_changepoint": bool(record["metrics"][metric_name]["percentage_change"]),
-            "percentage_change": record["metrics"][metric_name]["percentage_change"],
+            "percentage_change": f"{record['metrics'][metric_name]['percentage_change']:.2f}%",
```
sjug, line 851: This changes percentage_change from a numeric value into a string containing both the percentage and confidence label. Consumers parsing this column numerically will break. Please preserve the percentage column and add a separate Confidence column, matching the summary table added by this PR.

orion/utils.py
```diff
@@ -848,8 +848,14 @@ def create_record(record):
             ),
             metric_name: record["metrics"][metric_name]["value"],
             "is_changepoint": bool(record["metrics"][metric_name]["percentage_change"]),
```
sjug, line 850: This table reports one metric at a time, but the JSON flag only says that some metric changed on the row. Percentage change is also not a reliable substitute. Please store is_changepoint inside each metric object and use that per-metric value here.

docs/usage.md
```diff
@@ -285,6 +285,114 @@ Today is 27 Aug:
...
+
+### Standalone Reports
+
+When generating reports from JSON files with `--report`, confidence data is propagated from the JSON into the summary tables automatically — no additional flags needed.
+
+### Interpreting Results
+
+Use confidence indicators to triage changepoints:
+
+1. **Likely real (large/moderate shift)** — investigate immediately; this is a genuine regression with meaningful impact
```
sjug: These statistics can show evidence of a metric shift, but they cannot prove that product code caused a regression. Environment, workload, or measurement changes could also explain it. Please use wording such as "evidence of a meaningful metric shift; investigate the cause."

orion/run_test.py
```diff
@@ -248,6 +249,9 @@ def analyze(test, kwargs, is_pull=False):
 
     _, change_points_by_metric = algorithm.get_analysis_results()
     regression_flag = algorithm.regression_flag
+    confidence_by_metric = compute_confidence(
```
sjug: Confidence is recalculated in four branches, making it easy for a future path to return stale results. Please track the selected raw analysis-window dataframe and changepoints, then calculate confidence once before creating AnalysisResult. Do not use final_algorithm.dataframe directly because CMR has already collapsed it into two aggregate rows.

orion/pipeline/formatters/json_formatter.py
```diff
@@ -41,6 +41,12 @@ def format(self, data: AnalysisResult) -> dict:
                     "percentage_change"
                 ] = percentage_change
                 dataframe_json[index]["is_changepoint"] = True
+                confidences = data.confidence_by_metric.get(key, [])
+                cp_idx = value.index(change_point)
```
sjug: value.index(change_point) searches the list again during every iteration. Please use enumerate(value) and use that position directly. This avoids repeated searches and keeps confidence aligned if two changepoints compare equal.

orion/pipeline/formatters/base.py
```diff
@@ -128,6 +136,19 @@ def extract_regression_data(self, data: AnalysisResult) -> list:
 
         return regression_data
 
+    @staticmethod
+    def _add_confidence(data, metric, cp, metric_entry):
+        """Add confidence data to a metric entry if available."""
+        confidences = data.confidence_by_metric.get(metric, [])
+        cps = data.change_points_by_metric.get(metric, [])
+        try:
+            cp_idx = cps.index(cp)
```
sjug: The call sites are already inside the changepoint loop, so they can pass the changepoint's position into _add_confidence. That avoids searching the list again and removes ambiguity if two changepoints compare equal.

test-local.bats
```diff
@@ -138,7 +138,7 @@ setup() {
     exit 1
   fi
 
-  changepoint=$(echo '404.549 | https://prow.ci/2013174937652563968 | -- changepoint')
+  changepoint=$(echo '404.55% (Likely real [9.51] (large shift [0.00])) | https://prow.ci/2013174937652563968 | -- changepoint')
```
sjug: This assertion depends on exact SciPy output and rounding, so harmless numerical or test-data changes could break the integration test. Please match the stable label or output structure instead of the complete calculated value. The other new confidence assertions should be relaxed similarly.

orion/tests/test_confidence.py
```diff
@@ -0,0 +1,261 @@
...
+        after = np.array([200.0, 201.0])
+        result = _compute_stats(before, after)
+        assert result.sufficient_data is False
+
+    def test_zero_std_different_means(self):
+        before = np.array([100.0, 100.0, 100.0])
+        after = np.array([200.0, 200.0, 200.0])
+        result = _compute_stats(before, after)
+        assert result.sufficient_data is True
+        assert result.cohens_d == float("inf")
```
sjug: This test makes infinite Cohen's d the expected behavior, but that value produces non-standard JSON when serialized. When the degenerate-variance handling is fixed, please update this test to expect nullable statistics and an explanatory reason instead.

orion/tests/test_confidence.py
```diff
@@ -0,0 +1,261 @@
...
+            sample_size_after=1,
+        )
+        assert result.sufficient_data is False
+        assert result.p_value is None
+
+
+class TestGetSegments:
+    def test_edivisive_splits_at_index(self):
+        data = np.array([1.0, 2.0, 3.0, 10.0, 11.0])
+        before, after = _get_segments(cnsts.EDIVISIVE, data, 3)
```
sjug, line 119: This test locks in the global prefix/suffix comparison. When changepoints use their local neighboring segments, please update this test and add a multiple-changepoint reversal case so a later shift cannot hide an earlier one.

orion/confidence.py
```diff
@@ -0,0 +1,167 @@
...
+            std_after=float(np.std(after, ddof=1)) if n_after > 1 else None,
+        )
+
+    _, p_value = stats.ttest_ind(before, after, equal_var=False)
+
+    mean_before = np.mean(before)
+    mean_after = np.mean(after)
+    std_before = np.std(before, ddof=1)
+    std_after = np.std(after, ddof=1)
+    pooled_std = math.sqrt(
```
sjug, line 107: Welch's test allows the groups to have different variances, while this pooled-standard-deviation Cohen's d assumes a shared variance. This is common in practice, but please document the choice or consider an effect size designed for unequal variances.

sjug (changes requested):
Requesting changes mainly because of two issues in how confidence is calculated: the current approach can make random noise look like a real change, and can label a real change as noise. Details are in the inline comments, along with smaller suggestions on output format, code structure, and tests.

This PR also removes roughly 600 lines from test.bats and adds an unrelated pylint suppression in test_matcher_batch.py. The deleted scenarios remain in test-local.bats, so coverage is preserved, but these cleanups should be acknowledged in the PR description or moved to a separate change.

jtaleric:
/lgtm
tested locally. 🚀

orion/confidence.py
jtaleric: > Otava's existing window statistics may be reusable

that would be nice! are there suggestions on how we could adjust the existing windowing?

orion/utils.py
jtaleric: I would hope a user would simply use `jq` w/ `-o json`  instead of parsing the human readable test.. that would be terrible.

sjug: Fair point. If the table is human-readable output rather than a stable interface, JSON is the supported contract and compatibility is not a blocker. I would still prefer a separate Confidence column: it is easier to read and matches the summary table this PR adds. Happy to treat this as a non-blocking suggestion.

orion/confidence.py
sjug: Yes. For the Otava changepoints, I would reuse the local before/after statistics already attached to ChangePoint.stats, or preserve the exact segment boundaries and sample sizes Otava used before Orion filters changepoints. That fixes the current prefix/suffix problem. We should not derive the boundaries from the filtered changepoint list because a neighboring boundary may already have been removed.

This only fixes the comparison window. Otava's p-value is also involved in selecting the changepoint, so it is not independently calibrated. In the short term, I would report the local means, standard deviations, and effect size as descriptive context without presenting the p-value or interval as calibrated confidence. Fully calibrated inference would require rerunning the complete detector-and-selection process on resampled data that contains no real change, or using held-out runs.

jtaleric: changing the underlying algo could introduce new regressions in the pipeline, right? where the current implementation will categorize the regressions likelihood.

sjug: No algorithm change needed. The suggestion is to reuse statistics Otava already computes internally when it finds each changepoint, purely for the confidence annotation. Detection results would be identical.

orion/confidence.py
sjug: The neighboring-window change is directionally right, but cp_indices is built from cps after Orion filters changepoints by direction, threshold, ACK state, and correlation. If a recovery changepoint is filtered out, its boundary disappears and the original dilution problem returns.

I reproduced this with a 0 -> +10 -> -10 series. With only the reported first changepoint, the comparison gives d=0.002 and p=0.992. With both detector boundaries, it gives d=120.65 and p=5.4e-69.

Please preserve the raw detector boundaries or local statistics before Orion filters the changepoints, and use those raw boundaries for the comparison windows.

orion/run_test.py
sjug: Consolidating this call is good, but it now uses final_algorithm.dataframe, which is the CMR caveat mentioned in the original comment. cmr.py:40 has already collapsed that dataframe to two aggregate rows.

In a four-run example, this changes the reported context from n=(3,1), including the baseline standard deviation, to n=(1,1) with both standard deviations unavailable. Please retain the selected raw analysis-window dataframe separately and pass that dataframe to compute_confidence.

orion/utils.py
sjug: This remains unresolved. The table is per metric, but bool(percentage_change) is not an explicit per-metric changepoint marker. The JSON row flag cannot replace it because that flag only says that some metric changed on the row.

Please add an is_changepoint field to each metric object and use it here. I consider this non-blocking, but it would keep the table semantics explicit and avoid inferring state from a display value.

orion/tests/test_confidence.py
```diff
@@ -0,0 +1,341 @@
...
+        ]
+        assert sizes == [(3, 2), (2, 5)]
+
+    def test_reversal_detected_independently(self):
+        df = pd.DataFrame({
+            "cpu": [10.0, 10.0, 10.0, 10.0, 10.0,
+                    20.0, 20.0, 20.0, 20.0, 20.0,
+                    10.0, 10.0, 10.0, 10.0, 10.0],
+        })
+        cps = {"cpu": [_make_cp("cpu", 5), _make_cp("cpu", 10)]}
```
sjug, line 252: This test supplies both changepoints to compute_confidence, so it does not cover the failure where Orion filters out the recovery changepoint before confidence is computed. Please add a case where the raw detector boundaries contain both indices but the reported changepoints contain only the first. The first comparison should still stop at the filtered recovery boundary.

orion/confidence.py
```diff
@@ -0,0 +1,209 @@
...
+        ((n_before - 1) * std_before ** 2 + (n_after - 1) * std_after ** 2)
+        / (n_before + n_after - 2)
+    )
+
+    if pooled_std == 0 and mean_before != mean_after:
+        p_value = None
+        cohens_d = None
+    elif pooled_std == 0:
+        _, p_value = stats.ttest_ind(before, after, equal_var=False)
```
sjug, line 116: When both segments are constant and equal, this still calls SciPy and emits a precision-loss warning in the focused tests. The result is already known: p=1.0 and d=0.0. Please return those values directly so production logs remain clean.

sjug:
Thanks for the follow-up. Most of the original presentation, serialization, and maintainability concerns are addressed. I am keeping changes requested for the two original statistical blockers: post-selection calibration and comparison boundaries lost during changepoint filtering. The CMR raw-data issue also needs correction.

Before merge, please also refresh the PR description. It still shows the old "Likely real"/"Noise" labels and old JUnit layout, and it does not mention the test.bats deduplication or unrelated pylint suppression.

afcollins (commented):
/lgtm

I like it. Improves our understanding of identified changepoints.

@sjug I see a lot of comments from you. Are they required or nits?

The is_changepoint reliability seems rather serious.

sjug:
@afcollins Requested changes, not comments. As I summarized in the comment prior to yours.

jtaleric:
@chentex Curious when we think you can address @sjug's feedback? Thanks!

afcollins (commented):
I'm happy with the [output], but will leave approval to sjug

[1]: https://privatebin.corp.redhat.com/?b87a7ef64c5a52eb#BfTHqDapcoMSR3NKNRNgaCGFsXJ4TP2XG8fSNmZoRNYm

orion/algorithms/cmr/cmr.py
```diff
@@ -112,15 +117,17 @@ def combine_and_average_runs(self, dataFrame: pd.DataFrame):
...
-            if isinstance(dF.loc[0, column], (numpy.float64, numpy.int64)):
-                mean = dF[column].mean()
-                data2[column] = [mean]
-            else:
-                column_list = dF[column].tolist()
-                # Convert each item to string to handle lists, UUIDs, and other non-string types
-                non_numeric_joined_list = ','.join(str(item) for item in column_list)
-                data2[column] = [non_numeric_joined_list]
+            try:
+                numeric_col = pd.to_numeric(dF[column])
```
sjug: This converts every numeric-looking column before averaging, including metadata. For example, with versions 4.17, 4.18, 4.19, and 4.20, the collapsed baseline reports 4.18 as the previous version. That value is an average of the first three versions; the actual preceding run is 4.19. This can send an incorrect version to reports, Sippy searches, and GitHub change context. Please average only configured metrics or explicitly numeric data fields, and preserve version, UUID, and other metadata without numeric coercion.

docs/usage.md
```diff
@@ -285,6 +285,147 @@ Today is 27 Aug:
...
+
+These produce a human-readable label shown in the "Affected Metrics" summary table and embedded in JSON output.
+
+### Label Format
+
+Labels are driven by Cohen's d effect size thresholds (Cohen 1988: 0.2 small, 0.5 medium, 0.8 large). The p-value is included as descriptive context only — it is **not** calibrated for post-selection inference (the changepoint was already selected from this same data):
+
+| Label | Meaning |
+|-------|---------|
+| `Large shift (d=1.20, p=0.001)` | d >= 0.8 — strong evidence of a meaningful metric shift |
```
sjug: Non-blocking suggestion: Detector selection can also inflate the observed Cohen's d, so d >= 0.8 describes a large observed contrast but is not independent "strong evidence." Please replace "strong evidence" with wording such as "large observed standardized contrast" here, in the triage guidance, and in the PR description.

orion/confidence.py
```diff
@@ -0,0 +1,211 @@
...
+    )
+
+    if pooled_std == 0 and mean_before != mean_after:
+        p_value = None
+        cohens_d = None
+    elif pooled_std == 0:
+        p_value = 1.0
+        cohens_d = 0.0
+    else:
+        _, p_value = stats.ttest_ind(before, after, equal_var=False)
```
sjug, line 127: Another non-blocking suggestion: The equal-constant case is fixed, but SciPy still emits the precision-loss warning when only one segment is constant. I reproduced this with raw detector boundaries supplied, using a constant baseline and a slightly variable after segment. Please check for zero variance in either segment before calling SciPy. When only one segment has zero variance, the Welch comparison is still defined and can be calculated directly from the means, variances, and sample sizes already available here.

sjug (changes requested):
The previous review findings are addressed. This follow-up found one new blocking issue: CMR now averages numeric-looking metadata, which can fabricate a previous version and affect downstream context. The other two comments are non-blocking suggestions.

sjug (approved):
Thanks for working through all of this, @chentex. Confidence values are now presented descriptively, comparison windows use the detector's raw boundaries, CMR preserves real metadata and raw baseline context, and the earlier serialization and formatting issues are all addressed. Approving.

PR #452, opened 2026-08-03
Add doc blurb about using fan_out for more summary statistics

afcollins:
## Type of change

- [x] Documentation Update

## Description

Adds a note about how to use fan_out to collect more stats on a metric.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -587,6 +587,33 @@ This expands into three metrics: `apiserverCPU`, `multusCPU`, and `ovnCPU`, each
 4. Each expanded metric is a deep copy — mutations to one do not affect others
 5. The `fan_out` key is removed after expansion
 
+#### Example: Retrieve more summary statistics with less duplication
+
+Say you want to collect more than a single stastic about a metric, like mean, but also min, max. You can use `fan_out` to do this compactly:
+
+```
+tests:
+  - name: payload-cluster-density-v2
+    metadata:
+      ...
+    metrics:
+      - name: nodeMajorFaults-${ag}
+        metricName.keyword: nodeMajorFaults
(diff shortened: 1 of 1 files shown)


afcollins:
I noticed this actually uncovered a bug or unexpected behavior, so I will update or raise a new PR with the fix.

docs/configuration.md
```diff
@@ -582,6 +582,33 @@ This expands into three metrics: `apiserverCPU`, `multusCPU`, and `ovnCPU`, each
...
+          agg_type: ${ag}
+        fan_out:
+          - ag: avg
+          - ag: min
+          - ag: max
+```
+
+This will return data that looks like the following table output:
+```
+time uuid ocpVersion nodeMajorFaults-avg_avg nodeMajorFaults-min_min nodeMajorFaults-max_max
```
afcollins, line 608: The duplication is ugly, but required or else we only get the last result metric.

I have a plan for a fix, but haven't prepared the fix yet, since this is a simple workaround.

docs/configuration.md
```diff
@@ -582,6 +582,33 @@ This expands into three metrics: `apiserverCPU`, `multusCPU`, and `ovnCPU`, each
...
+    metadata:
+      ...
+    metrics:
+      - name: nodeMajorFaults-${ag}
+        metricName.keyword: nodeMajorFaults
+        metric_of_interest: value
+        agg:
+          agg_type: ${ag}
+        fan_out:
+          - ag: avg
```
afcollins, line 601: If the key is `agg` or `agg_type` it can cause unexpected errors. Maybe we could validate the fan_out block for known keywords?

chentex: It would be good to have a config validation.

PR #453, opened 2026-08-04
Adding node page fault as a metric

vishnuchalla:
## Type of change

- [x] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adding `nodeMajorFaults` as a metric to the config files.

- Closes # https://redhat-internal.slack.com/archives/CNKHL6076/p1785525895313329

## Testing
Tested and verified in local using
```
  orion \
    --hunter-analyze \
    --node-count \
    --sippy-pr-search \
    --viz \
    --debug \
    --config "examples/trt-external-payload-cluster-density.yaml" \
    --lookback "30d" \
    --es-server "$ES_SERVER" \
    --benchmark-index "$INDEX" \
    --metadata-index "$INDEX" \
    --input-vars '{"version": "5.0"}'
```

### DIFF
--- a/examples/metrics/chaos-scenarios-metrics.yaml
+++ b/examples/metrics/chaos-scenarios-metrics.yaml
@@ -460,3 +460,43 @@
   metric_of_interest: value
   agg:
     agg_type: avg
+
+- name: max-mutating-apicalls-latency-${ag}
+  metricName.keyword: max-mutating-apicalls-latency
+  metric_of_interest: value
+  agg:
+    agg_type: ${ag}
+  direction: 1
+  fan_out:
+    - ag: avg
+    - ag: max
+
+- name: max-ro-apicalls-latency-${ag}
+  metricName.keyword: max-ro-apicalls-latency
+  metric_of_interest: value
+  agg:
+    agg_type: ${ag}
+  direction: 1
+  fan_out:
+    - ag: avg
+    - ag: max
+
+- name: 99thEtcdCompaction-${ag}
+  metricName.keyword: 99thEtcdCompaction
+  metric_of_interest: value
+  agg:
+    agg_type: ${ag}
+  direction: 1
+  fan_out:
+    - ag: sum
+    - ag: max
+
+- name: nodeMajorFaults-${ag}
+  metricName.keyword: nodeMajorFaults
+  metric_of_interest: value
+  agg:
+    agg_type: ${ag}
+  direction: 1
+  fan_out:
+    - ag: max
+    - ag: sum
--- a/examples/crd-scale.yaml
+++ b/examples/crd-scale.yaml
@@ -117,3 +117,43 @@ tests:
           agg_type: avg
         direction: 1
         threshold: 10
+
+      - name: max-mutating-apicalls-latency-${ag}
+        metricName.keyword: max-mutating-apicalls-latency
+        metric_of_interest: value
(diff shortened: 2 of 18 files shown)


examples/metrics/chaos-scenarios-metrics.yaml
```diff
@@ -460,3 +460,14 @@
...
+
+- name: nodeMajorFaults-${ag}
+  metricName.keyword: nodeMajorFaults
+  metric_of_interest: value
+  agg:
+    agg_type: ${ag}
+  direction: 1
+  fan_out:
+    - ag: avg
+    - ag: min
```
afcollins: min is not needed, it seems to always be 0. we can remove it

Also add the ro-apilatency metric too please

vishnuchalla: Added ro-apilatency as well. 

> min is not needed, it seems to always be 0. we can remove it

No, this doesn't seem to be true all the time. I have seen min value being populated in etcd sharding tests I performed.

rsevilla87: I don't think we should the check the min value of this metric.

vishnuchalla: > I don't think we should the check the min value of this metric.

Why not?

Node_vm_pagefault_sharded
Node_vm_pagefault

afcollins: The metric `nodeMajorFaults` [records](https://github.com/kube-burner/kube-burner-ocp/blob/main/cmd/config/metrics-profiles/metrics.yml#L215-L216) the rate of the metric `node_vmstat_pgmajfault`.

We don't need the min because it should always be 0, and if it isn't, then we are already capturing the rate of major faults under avg and max.

We would only want min if we were recording the raw counter value, so we could calculate the delta in the counter value across the run.

afcollins:
So many files modified. We can't use inheritance for this?

examples/metrics/chaos-scenarios-metrics.yaml
```diff
@@ -460,3 +460,46 @@
...
+
+- name: 99thEtcdCompaction-${ag}
+  metricName.keyword: 99thEtcdCompaction
+  metric_of_interest: value
+  agg:
+    agg_type: ${ag}
+  direction: 1
+  fan_out:
+    - ag: sum
+    - ag: avg
```
afcollins: I know I suggested it, but if we have sum, we probably also don't need avg.

vishnuchalla:
> So many files modified. We can't use inheritance for this?

That is out of scope for this PR. We need to do a different PR that does all changes related to inheritance.

examples/metrics/chaos-scenarios-metrics.yaml
vishnuchalla: Udpated.

examples/metrics/chaos-scenarios-metrics.yaml
afcollins: max and sum are most useful of these for nodeMajorFaults. You can remove min and avg.

vishnuchalla: Okay, removed and updated.

PR #455, opened 2026-08-04
Move imports in main so that '--help' returns much faster

afcollins:
## Type of change

- [x] Optimization

## Description

`orion --help` takes too long for what it is meant to do. It should respond instantly.

The source of the delay is loading all the modules.

This moves the modules to be lazily imported on the functions where they are used.

## Testing
```
$ time orion --help
Usage: orion [OPTIONS]
...
orion --help  0.11s user 0.04s system 85% cpu 0.177 total
```

Tests pass:
```
$ make test                                                                                                                                                                                                                                                                  
...
collected 254 items                                                                                                                                                                                                                                                         
...
254 passed, 4 warnings in 2.03s
```


jtaleric:
yeah, i figured lint would have a field day with this. 

not opposed to the change, but we might need to ignore some more linting.

afcollins:
Yeah, good call. I have a less-invasive idea of simply putting the --help output in the readme, verbatim. Then we could keep it up to date in an automated way.

PR #456, opened 2026-08-05
Check for duplicate metrics and warn the user

afcollins:
Prior behavior will just throw an uncaught exception

## Type of change

- [x] Bug fix

## Description

If there are two metrics with the same name, then an exception can be thrown.

Instead, we can check if there are duplicate metric names and log a warning.


jtaleric (commented):
/lgtm

PR #457, opened 2026-08-06
Move pr column to the end of csv

afcollins:
## Type of change

- [x] Refactor

## Description

Since the pr column can have many commas in it
we move it to the end so we can cut it from
a csv and process it easily with other tools.


jtaleric (approved):
/lgtm

PR #472, opened 2026-09-04
Adds new dry-run metric type

### LINKED ISSUE #426: [RFE] dry-run metrics
## What we need

Currently, when Orion evaluates a metric and detects a regression or a threshold breach, it automatically triggers its default alerting behaviors, such as failing the execution (exiting with a non-zero return code) or creating a Jira ticket.

However, there are many scenarios where we want to introduce new, experimental metrics or tune existing thresholds without the risk of breaking pipelines or generating noisy Jira tickets. Right now, there is no way to "canary" or test a metric definition in a safe, non-blocking mode.

## Describe the solution you'd like

Introduce a **dryRun** parameter for metric configurations. When a metric is marked with `dryRun: true`, Orion should still execute the comparison, evaluate the threshold, and log the results as usual. However, it must explicitly suppress any disruptive failure actions. Specifically, it should:

- Prevent the creation of a Jira ticket
- Prevent the execution from failing (ensuring the process exits with rc = 0).

## How

```yaml
- name: experimentalMetric
  metricName.keyword: myExperimentalMetric
  metric_of_interest: value
  direction: 1
  threshold: 10
  dryRun: true
```

chentex:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adds new dry-run metric type

- Related Issue #426 
- Closes #426 

## Testing

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring metrics as dry runs.
  - Dry-run metrics continue to be analyzed and reported, including visualizations and regression details.
  - Dry-run metrics do not create JIRA issues, affect exit codes, or undergo acknowledgment filtering.

- **Bug Fixes**
  - Added validation to prevent dry-run metrics from being used as correlation targets.
  - Dry-run settings now default to disabled when unspecified.

- **Documentation**
  - Documented dry-run metric configuration and behavior.

### DIFF
--- a/orion/pipeline/formatters/base.py
+++ b/orion/pipeline/formatters/base.py
@@ -65,8 +65,8 @@ class BaseFormatter(ABC):
                                 "percentage_change": percentage_change,
                                 "labels": data.metrics_config[metric].get(
                                     "labels") or [],
-                                "dryRun": data.metrics_config[metric].get(
-                                    "dryRun", False),
+                                "dry_run": data.metrics_config[metric].get(
+                                    "dry_run", False),
                             })
                             self._add_confidence(
                                 data, metric, cp_idx,
@@ -96,7 +96,7 @@ class BaseFormatter(ABC):
                         "value": row.get(metric),
                         "percentage_change": percentage_change,
                         "labels": data.metrics_config[metric].get("labels") or [],
-                        "dryRun": data.metrics_config[metric].get("dryRun", False),
+                        "dry_run": data.metrics_config[metric].get("dry_run", False),
                     }],
                     "prs": [],
                     "github_context": None,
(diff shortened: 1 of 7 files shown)


orion/pipeline/formatters/base.py
```diff
@@ -94,6 +96,7 @@ def extract_regression_data(self, data: AnalysisResult) -> list:
                         "value": row.get(metric),
                         "percentage_change": percentage_change,
                         "labels": data.metrics_config[metric].get("labels") or [],
+                        "dryRun": data.metrics_config[metric].get("dryRun", False),
```
rsevilla87: I think it's better to use snake_case naming convention here

rsevilla87 (commented):
/lgtm

PR #475, opened 2026-09-09
fix: exclude detected anomalies from moving average window in Isolation Forest

### LINKED ISSUE #473: Anomaly detection moving average includes flagged anomalies, causing cascading missed detections
**Description**
The Isolation Forest anomaly detection algorithm does not exclude detected anomalies from the rolling moving average window. This means a flagged anomalous run still contributes to the baseline used to evaluate subsequent runs.

In the case of a single anomaly, the poisoned value remains in the moving average window for the next window_size - 1 runs, dragging the baseline toward the anomalous value. During this period, moderate regressions can appear "within normal range" and go undetected.

In the case of multiple consecutive anomalies, the moving average converges further toward the anomalous values with each additional bad run. Once the number of consecutive anomalies equals or exceeds the window size, the baseline becomes fully poisoned — the percentage change for any new anomalous run approaches 0%, falls below the `min_anomaly_percent` threshold, and is silently dropped. From that point on, no further anomalies can be detected until enough healthy runs flush the bad values out of the window.

**Expected Behavior**
Once a run is flagged as anomalous, it should be excluded from the moving average window used to evaluate subsequent runs. Each build should be compared against the last N healthy runs, not the last N runs regardless of status.

**Suggested Fix**
After the Isolation Forest produces predictions, compute the moving average using only non-anomalous data points. This ensures that anomalous data points never pollute the baseline for subsequent anomaly checks.

shashankkestwal:
The rolling moving average used to calculate percentage change for anomaly detection included flagged anomalous runs. This caused the baseline to converge toward anomalous values when consecutive bad runs occurred, making subsequent anomalies invisible (0% change).

Replace the single-pass rolling mean with a clean version that sets anomalous rows to NaN before computing the window average, and forward-fills to preserve the last known healthy baseline when the entire window is anomalous.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

The rolling moving average used to calculate percentage change for anomaly detection included flagged anomalous runs. This caused the baseline to converge toward anomalous values when consecutive bad runs occurred, making subsequent anomalies invisible (0% change).

Replace the single-pass rolling mean with a clean version that sets anomalous rows to NaN before computing the window average, and forward-fills to preserve the last known healthy baseline when the entire window is anomalous.

- Related Issue #
- Closes #473

## Testing

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved weighted moving-average calculations by excluding detected anomaly values from the calculation.
  * Moving averages now continue across anomalous data points and retain valid results when limited data is available.
  * Prevented anomalous zero values from distorting the moving average for subsequent runs.

* **Tests**
  * Added coverage verifying that anomalous runs are correctly identified without affecting later moving-average results.


chentex:
Thanks for the PR @shashankkestwal , can you review coderabbits comments and review the integration tests, I believe this breaks the current behavior and they need updating.

chentex:
@shashankkestwal also the commits need to be signed

chentex:
@shashankkestwal Unit test are failing

orion/algorithms/isolationforest/isolationForest.py
```diff
@@ -44,7 +44,8 @@ def _analyze(self):
 
         # Calculate moving average for each metric
         window_size = (5 if self.options.get("anomaly_window",None) is None else int(self.options.get("anomaly_window",None)))
-        moving_averages = dataframe_with_metrics.rolling(window=window_size).mean()
+        healthy_metrics = dataframe_with_metrics.loc[dataframe["is_anomaly"].ne(-1)]
```
vishnuchalla, line 47: The whole purpose of isolation forest was to detect anomalies. if we discard them, doesn't that defeat the whole purpose?

shashankkestwal: Good question! The anomalies are not being discarded - Isolation Forest still detects them (line 39-40). What's being excluded is the anomalies from the moving average calculation used as the baseline for comparison (line 47-48).

vishnuchalla: @shashankkestwal wdyt?

shashankkestwal: In IsolationForest, runs that are statistical outliers get flagged as anomalies. The current fix ensures these flagged runs are excluded from the moving average window — so the baseline stays grounded in healthy run values and doesn't drift toward anomalous ones.
Without this, a single bad run (say e.g. rps=0) would pull the rolling mean down, making the next genuinely regressed run look acceptable by comparison.





**As a potential alternative approach**: Instead of relying on IsolationForest to decide what gets excluded from the baseline, we could compute the moving average sequentially at runtime — only including runs that fall within the threshold in the window. Runs that deviate beyond the threshold would be flagged and skipped from future baseline calculations automatically.

vishnuchalla (approved):
/lgtm

